PR #7, opened 2019-01-20
WIP Remove RMQ, Add API POST request method.

### LINKED ISSUE #5: Remove RMQ dependency
Since we're no longer using RMQ in the Django rewrite, this project needs to be rewritten slightly to rely on a different method for receiving the eval requests. An API seems like the natural thing to replace it with.

scragly:
Closes #5 

Related `bot` PR:
https://github.com/python-discord/bot/pull/289

This adds the changes necessary to remove the reliance on RMQ, instead adding a new POST endpoint to receive input, process, then return the result.

The bot has a related PR coming that implements the required changes in order for it to request results from snekbox without RMQ.

By default, it will run on `http://localhost:8060`, with the main index being a test page to submit input and see it's result, effectively merging `snekweb` into `snekbox`.

The API endpoint to send a POST request is `/input`, making the full default URI the bot has to be configured with `http://localhost:8060/input`.

The request `Content-Type` has to be set to `application/json`, and the json data should have the following structure:
```json
{
  "code": "print(1)"
}
```
The returned json data will be structured as follows:
```json
{
  "input": "print(1)",
  "output": "1"
}
```

There may be considerations that I have not been able to take into account regarding how production is setup, so let me know if there's any adjustments required.

If you want to test this PR, you must pull the `snekbox/rmq_removal` branch instead of the default, run in snekbox project directory:
```
pipenv run build
docker-compose up -d
```
Which then you can check it's live on `http://localhost:8060`.


Pipfile
```diff
@@ -4,15 +4,12 @@ verify_ssl = true
...
 [packages]
-pika = "*"
 docker = "*"
-
-[dev-packages]
 flask = "*"
-flask-sockets = "*"
-gevent = "==1.2.2"
-gevent-websocket = "*"
+gevent = "*"
```
MarkKoz: Is this still being used for anything? I can't find anything.

Inveracity:
looks good to me 👍

PR #17, opened 2019-03-29
Restructure Docker Images

### LINKED ISSUE #8: Python 3.7 migration
~~Adjust dependency management to use poetry to be in line with other projects in the org, and~~

Up the Python version used to 3.7.

MarkKoz:
This has the aim of reducing build time by splitting up the Docker images, which allows for the images to be built only when needed on an individual basis. I'm hoping that if this PR is well received, it will set a precedent for how CI and deployment is done on the rest of our repositories. 

## Docker Images
Images were split into 3:

#### Base Image
This image contains NsJail as well and the Python environment and pipenv. This image should very rarely need to be built. It takes a bit over a minute to build.

I made use of a multi-stage build. The first stage is tasked with building NsJail from source, using the latest commit on their repository. This first stage uses a plain old Alpine image. The built NsJail binary is copied to the second stage, which is what ends up in the final image. The second image uses an Alpine image with Python 3.7.3 (closes #8). 

#### Venv Image
This image copies the the Pipfiles and creates the virtual environment. The idea is that this image will only be built when either Pipfile changes, which will make total build times shorter most of the time. This image takes about 30 seconds to build.

#### Final Image
This image is very simple: it just contains the source code of the project. Thus, it takes merely 4 seconds to build. This is great because this image will be the only one that is _always _ built (except when in a PR).

## CI
All images are built by default. A bash script determines whether images shouldn't be built, using `git diff` to compare the head and the latest commit of the most recently successful build. In order to retrieve such commit, the Azure DevOps REST API is employed. PRs will fall back to using the head of the target branch if no previous successful build is found. Regardless of the diff's results, images will be built if they do not yet exist (i.e. they have never been pushed to Dockerhub before).

When their Dockerfiles change, images are also built in PRs because I feel it's better to catch issues with image builds _before_ merging rather than after. Editing Dockerfiles is an infrequent operation and it wasn't much harm to let the images be built. However, dependency changes (i.e. changes to the Pipfile) do not trigger the venv image's build in a PR.

The final image is never built in a PR, and always built outside of one. I didn't feel it was worth the effort because the Dockerfile should never need to be edited. If it does, given that the changes respect the image's purpose and actually belong in this image, I don't see how it could be changed in a manner that would end up breaking the build.

## Concerns
1. NsJail is built from the latest commit. There is uncertainty to how stable master is. It would be easy to build from a release instead. So far, most of the commits since the latest release seem to just be refactoring and miscellaneous things. Presumably, any major bug fix would prompt for a new release.
2. ~~I am leaning towards making dependency changes also build the venv image in PRs. Frankly, dependency changes aren't too common either. Furthermore, the fact that the virtual environment creation step in the Lint job succeeds cannot be relied upon because the Ubuntu agent used is a significantly different environment that comes with a lot more packages out of the box.~~ Made this change.
3. I would have liked for images to be built even if the "Check Changed Files" would fail; I don't know how robust that bash script is or how reliable the REST API is. However, I couldn't figure out a way to cleanly implement that because of the reliance on `condition: succeeded()`, which only works if _all_ previous steps for the job succeeded. The problem with `succeededOrFailed()` is that it would permit more than one step to fail, where as ideally, only the "Check Changed Files" step would be permitted to fail.
4. Any suggestions on how to make `azure-pipelines.yml` neater, particularly the bash script in "Check Changed Files", would be appreciated. There are some long lines which I don't know how to split up.


MarkKoz:
@scragly Is this sort of what you had in mind for splitting up the images? Note I haven't done anything to the build pipeline yet.

PR #23, opened 2019-05-29
Replace Flask with Falcon

### LINKED ISSUE #14: Replace Flask
With #7 in the works, Flask seems overkill. I suggest we switch to something more light-weight like Bottle or Falcon. 

Falcon looks to be more purpose-built for our needs (i.e. solely an API framework), but writing code for it is a bit bulkier since it's class-based rather than decorator-based. Generally, I'd argue this is a positive. However, we only need a single endpoint so it's a bit much in this case. Falcon also has a nice API for use in writing tests.

Bottle is decorator-based and its code generally looks very similar to Flask's. It also has no dependencies so that's a plus.

MarkKoz:
Closes #14.

Todo:

- [x]  Document the API. What format should it have? Similar to the API docstrings on django rewrite?
- [x] Logging


sco1:
Documentation is still an open discussion organization-wide (https://github.com/python-discord/organisation/issues/35) since we haven't really yet had anything formally documented.

At a minimum, I think we should aim to cover:

- Relevant public APIs documented, ideally with examples
- Instructions for reliably getting things running in a dev environment

Having these as the base will ideally enable quicker refresh for new or returning contributors, along with expanding the pool of available reviewers for new PRs.

Last we spoke as an organization, these are intended to live on the site wiki once it's launched, but can live in the project's GH wiki until then.

gdude2002:
Are you sure you guys have investigated all the alternatives?

Not that Falcon is particularly bad, but I've used it myself for some projects and it has some really weird problems, mostly relating to cookie handling (which I guess you don't care about).

If you're going to use Falcon for an API, then you might as well use Hug, since it builds off it, right?

lemonsaurus:
Let's not overthink this. It's not important to choose the very best framework in the world for an API that will never have more than a single endpoint, it's just not worth the rework. I don't know if it was worth switching from Flask to Falcon either, but that work has already been done and I don't really mind it. Besides, Falcon is probably the better choice out of the two.

MarkKoz:
@sco1 I'll just follow the site's API docstring style for now. I can revisit it once we have finalised how we want things documented. Also, you have a good point on instructions. I was also planning on adding some docker provisions for local testing and development (since NsJail causes issues otherwise).

@gdude2002 I'm sticking to Falcon, but I can provide insight on why it was chosen if you're curious. I looked into alternatives before raising an issue with everyone else. In the end, only Falcon, Bottle, and Hug seemed like contenders. 
* Hug was dropped cause, while it has a simpler API (decorator-based) than Falcon, it's bulkier with more dependencies. The simpler API doesn't really matter anyway; they're just conveniences and we only have 1 endpoint. Though one could also argue the extra dependencies don't really matter either here.
* Bottle had no dependencies and had a decorator-based API, but its inactive development was a major concern. Another downside is that it didn't seem to have any built-in provisions for unit tests where as Falcon easily works with both unittest and pytest out of the box.

I brought up both Bottle and Falcon and it was decided Falcon is better.

PR #24, opened 2019-05-30
Improve NsJail

MarkKoz:
- [x] Enforce double quotes with flake8 plugin
- [x] Lint tests
    - [ ] Add CI step for linting tests
- [x] Fix pre-commit
- [x] Refactor NsJail
- [x] Improve logging
- [x] Provide return code in return value of `NsJail.python3()` 
    * Accomplish by editing and returning the `subprocess.CompletedProcess`?
- [ ] Refactor tests
- [ ] Improve test coverage
- [ ] Provisions for a Docker image for running tests locally


Pipfile
```diff
@@ -20,12 +20,19 @@ flake8-tidy-imports = "*"
...
 
 [scripts]
-lint = "flake8"
+lint = """
+python -c
+\"import subprocess, sys
+files = subprocess.check_output(('git', 'ls-files', '--', '*.py')).decode('UTF-8').strip()
+args = ['pre-commit', 'run', 'flake8', '--files'] + files.split('\\n')
+subprocess.run(args, stdout=sys.stdout, stderr=sys.stderr)\"
+"""
```
jchristgit: why would we run this only via the files tracked by git in a subprocess (python -c) spawned by another subprocess (pipenv) when we can just run `flake8`?

MarkKoz: Yeah I figured this may be controversial, but I explained it in the commit message:

> Pre-commit is a nice wrapper for this particularily because it allows
> both the snekbox and tests to be linted with a single command while
> clearly distinguishing their results.
> 
> The Python script is a cross-platform implementation of the following:
>     git ls-files -- '*.py' | xargs pre-commit run flake8 --files

MarkKoz: I should make it clear this was added because two `flake8` invocations would otherwise be required (one for snekbox and one for tests). However, upon looking into it more, turns out they did end up following through with a feature request for (`--per-file-ignores`) so I'll investigate that and see if it will remove the need for even having 2 `.flake8` files.

MarkKoz:
OK my rebases may have fucked the merge commit. Fuck.

MarkKoz:
I decided to do the testing changes in a separate PR for the sake of making this PR easier to review.

PR #33, opened 2019-06-30
CI Improvements

### LINKED ISSUE #25: Venv Image Being Built Without Building Base First
As seen in [this](https://dev.azure.com/python-discord/Python%20Discord/_build/results?buildId=2343) build.

### LINKED ISSUE #26: Can CI be configured to completely skip the build job?
Just an idea to explore. Spinning up an agent and checking out take up a little bit of time even when there are no images to build or push. The "Check Changed Files" step would be moved to the end of the lint job and hopefully it'd be possible to set the environment variables and propagate them to the build job.

MarkKoz:
The diff looks daunting because of refactoring that I did. I encourage you to look at [this diff](https://github.com/python-discord/snekbox/pull/33/files/7cedf0c308a74bcb59f65d7d90978d3c7b43bf85..2f864a14a3f4d49d11801af45946dbd81e3e343f) instead; it skips past the first two refactoring commits.

### Changes

Please see individual commits for more details. I put effort into ensuring the commits are organised and make sense.

#### Shell Script

The shell script which checks for diffs of Dockerfiles has been reworked:

* Moved to a separate file
* Follows better shell script conventions
* Should be far more robust in terms of catching and handling errors
* Uses `jq` instead of an inadequate regular expression

I've also simulated Azure API calls for builds on master. Everything seems to work and be accounted for, but of course I cannot perform an actual test of a build on master.

#### Job and Step Conditions

Fixed #25 and resolved #26. Comments have been added in the YAML file to explain the conditions for most of the tasks. Hopefully they will ease any confusion.

#### Docker Tasks

Instead of script tasks, Docker tasks are used where possible. The login has also been changed to use the v2 Docker Registry API (v1 is deprecated). Logging in works but I have obviously not tested pushing.

### Comments

There is a redundancy in building the base image in both jobs. I chose to do this because the more efficient solution would involve moving almost everything into a single job, which I felt is more disorganised. Thoughts on this would be appreciated.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -89,6 +89,15 @@ jobs:
 
   - job: build
     displayName: 'Build'
+    condition: >
+      and(
+        succeeded(),
+        or(
+          ne(variables['Build.Reason'], 'PullRequest'),
+          eq(coalesce(dependencies.test.outputs['check.BASE_CHANGED'], True), True),
+          eq(coalesce(dependencies.test.outputs['check.VENV_CHANGED'], True), True)
+        )
+      )
     dependsOn: test
 
     variables:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -91,9 +91,10 @@ jobs:
           testResultsFiles: '**/TEST-*.xml'
           testRunTitle: 'Test Results'
 
+      # Run report too because the XML report doesn't output to stdout.
       - script: |
           docker exec snekbox_test /bin/ash -c \
-            'pipenv run coverage xml'
+            'pipenv run /bin/ash -c "coverage report && coverage xml"'
         displayName: 'Generate Coverage Report'
 
       - task: PublishCodeCoverageResults@1
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -110,7 +110,10 @@ jobs:
           and(
             succeeded(),
             ne(variables.BASE_PULL, True),
-            eq(variables.BASE_CHANGED, True)
+            or(
+              eq(variables.BASE_CHANGED, True),
+              eq(variables.VENV_CHANGED, True)
+            )
           )
         inputs:
           command: build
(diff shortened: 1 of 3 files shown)


MarkKoz:
The v2 Docker tasks may rely on logging in with v2 after all... In order to use it, [this](https://docs.microsoft.com/en-us/azure/devops/pipelines/library/service-endpoints?view=azure-devops&tabs=yaml#sep-docreg) needs to be set up. Could just switch back to shell scripts if that can't be done.

azure-pipelines.yml
```diff
@@ -1,144 +1,227 @@
...
+            --volume "${PWD}":"${PWD}" \
+            --workdir "${PWD}"\
+            --entrypoint /bin/ash \
+            pythondiscord/snekbox-venv:dev
+        displayName: 'Start Container'
+
+      - script: |
+          docker exec snekbox_test /bin/ash -c \
+            'pipenv run lint --format junit-xml --output-file test-lint.xml'
+        displayName: 'Run Linter'
```
jchristgit: with `--format junit-xml` do you not remove the default output?

MarkKoz: Yes it does. I would like to have both but not sure how.

MarkKoz: The `--tee` option would make it print to stdout but it'd still be in the XML format. The only solution I can think of is to run flake8 twice with different options. Thoughts?

jchristgit: I think it would be better to just leave the JUnitXML reporting out. I also like it, but I think that relying on Azure to properly display the lint output is a bit unintuitive.

MarkKoz: I don't think it's too taxing to run the linting twice but I do admit I never really cared much for the fancy output for linting in particular. It would create separate entries for everything IIRC.

Testing seems to still output to stdout so that's fine. What about the coverage report, which doesn't output to stdout? The output on Azure is quite nice. Should the plain old `coverage report` be ran too in CI? I don't think it's taxing since it just reads the already made `.coverage` file.

jchristgit:
Thanks!

PR #35, opened 2019-08-05
Rewrite README

MarkKoz:
Let me know if you think anything is missing or unclear.

I've also done some miscellaneous fixes and improvements to the scripts.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -29,9 +29,9 @@ python_version = "3.7"
 [scripts]
 lint = "flake8"
 precommit = "pre-commit install"
-test = "scripts/dev.sh -c 'pipenv run coverage run -m unittest'"
+test = "sh scripts/dev.sh -c 'pipenv run coverage run -m unittest'"
 testb = """
-    scripts/dev.sh \
+    sh scripts/dev.sh \
         --build \
         --clean \
         -c 'pipenv run coverage run -m unittest'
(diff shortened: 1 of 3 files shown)


SebastiaanZ:
This looks good to me. The README is clear, the only thing I'm less sure about is how a dev set-up of Snekbox is supposed to interact with a dev set-up of the bot/site running: Does this work out of the box or do we need to set up something in addition to the steps mentioned in the README?

(That question probably shows my ignorance of the subject.)

MarkKoz:
This is a standalone program. The bot depends on snekbox, but snekbox does not depend on the bot. You could manually send POST requests to snekbox if you want. Do you think I should make this clearer in the README?

SebastiaanZ:
No, I don't think so; I don't think it directly related to Snekbox by itself.

I was mainly thinking about the case in which someone wants to set up a complete development environment locally for the bot (which includes site) with Snekbox. Is that possible? Or would that still be difficult? Regardless, the information on how to do that should probably live in the bot's documentation, not Snekbox.

MarkKoz:
Yes it's possible and simple. The instructions for setting up snekbox wouldn't really change at all if you needed to use it with the bot.

You did remind me that I should document the API endpoint though.

MarkKoz:
There's a substantial amount of text here https://github.com/python-discord/snekbox/blob/2d1a0ec4a9fa7cf0758e8cc75a42cf93c1d23ebc/snekbox/api/resources/eval.py#L38

Should I just post a link to it in the README or add it to the wiki? I'm not keen on copy pasting the whole thing into the README.

SebastiaanZ:
I think a link in the README should suffice, even if it's just a reference/link to the relevant lines in the file. I'd rather keep it DRY and link to the information, then copy-pasting it around. That also makes sure that the information is in sync, instead of having two copies floating around that could get out of sync with each other.

PR #44, opened 2019-09-21
Research 2: Electric Boogaloo

MarkKoz:
no peeking top secret testing for the government


lemonsaurus:
Does this really need to be open, @MarkKoz?

MarkKoz:
No I'll just open it or a new one if I need to test CI again.

PR #46, opened 2019-09-27
Update flake8-annotations dependency pinning

sco1:
flake8-annotations v1.1 was recently released, which includes a fix for incorrect parsing of nested functions.

This PR updates the dependency pinning, no linting changes required.


GhostofGoes:
Any idea why the pipeline is failing on typed_ast? It's definitely in Pipefile.lock

MarkKoz:
> 
> 
> Any idea why the pipeline is failing on typed_ast? It's definitely in Pipefile.lock

https://paste.pythondiscord.com/ebelenecex.txt

Manylinux wheels don't work on muslc-based distros such as Alpine. Therefore, it tries to build from source but cannot because gcc is not installed. Could be fixed but the images will probably be moved to python slim images like the other repos, which I think is Debian.

MarkKoz:
Closing because #50 will take care of this and needs to update it to an even later version.

PR #49, opened 2019-10-05
Use pipenv sync instead of pipenv --sync

makusu2:
The pipenv docs page certificate seems to be having issues at the moment, but `pipenv --sync` doesn't work for me (Ubuntu 18.04, pipenv version 2018.11.26, Python3.7.4, Gnome Terminal), as described in the README; what works for me, and what I generally see used instead, is `pipenv sync`. Could we change the README to reflect this?

Thanks! ^_^


makusu2:
The PyPi page isn't down, and it suggests `sync` instead of `--sync` https://pypi.org/project/pipenv/

PR #51, opened 2019-12-14
Add external libraries and update image

Akarys42:
Hi, this PR add the external libraries specified in #3 to the Pipfile, so they can be used during eval :
* numpy 
* pandas  
* sympy 
* fuzzywuzzy 
* python-dateutil 
* pendulum
* arrow 
* yarl 
* attrs 
* forbiddenfruit
* more-itertools
* networkx 
* bs4
* pyyaml 
* toml

@MarkKoz have also updated the docker base image to Debian Buster, so we can use gcc.


Akarys42:
Apparently, the container build is failing because of gcc missing. How do you think we should tackle this problem @scragly and @MarkKoz ?

Pipfile
```diff
@@ -7,6 +7,19 @@ name = "pypi"
...
+scipy = "*"
+pandas = "*"
+sympy = "*"
+fuzzywuzzy = "*"
+python-dateutil = "*"
+pendulum = "*"
+arrow = "*"
+yarl = "*"
+pyyaml = "*"
+toml = "*"
```
MarkKoz: I'm not sold on including toml & yaml libraries since there are no write permissions nor internet access.

Pipfile
```diff
@@ -7,6 +7,19 @@ name = "pypi"
 falcon = "~= 2.0.0"
 gunicorn = "~= 19.9"
 jsonschema = "~= 3.0"
+numpy = "*"
```
MarkKoz: All packages should have their versions pinned.

Akarys42: Is there a fats way to do it, or it needs to be done by hand? Also, I don't really see the point of pinning the version of external dependencies, since it isn't used in the snekbox code directly, and therefore can't break it. Plus, it will avoid us to manually update a package when a new version is out.

MarkKoz: > Is there a fast way to do it, or it needs to be done by hand?

Maybe `pip freeze` can help, otherwise by hand. 

> since it isn't used in the snekbox code directly, and therefore can't break it

Still would rather not have anything break even if it's one package.

> Plus, it will avoid us to manually update a package when a new version is out.

Fair point perhaps. However, I don't think this would happen often so I'm OK with manual updates. Also, either way, a rebuild will still need to be triggered.'

Let's see what anyone else has to say.

lemonsaurus: Let's just pin'em.

Akarys42:
Yes, sir!
Also, what about bs4, yaml and toml? Sedsarq said they’d like to have them,
to demonstrate basic usages with pasted file. Also, I think we could in a
near future add some basic HTML, yaml and toml files in read-only, for
demonstration purposes.

lemonsaurus:
I guess I don't see the harm in `bs4`, `toml` and `yaml` being added. they're all popular modules.

SebastiaanZ:
If people feel that they're going to be used with local examples (e.g., strings) for demonstration purposes, I see no harm in adding them. I think it will create some confusion since people will undoubtedly try to request external files, but they'll notice soon enough that you can't do that with Snekbox.

PR #55, opened 2019-12-29
Configure a proper chroot jail for NsJail

MarkKoz:
## Description

devfs and sysfs were problematic since they were being mounted as tmpfs, which is r/w. For example, the Python process could write to cgroups. Now, only what is needed to run Python gets mounted. This boils down to the venv itself and some shared libraries Python needs.

* Use a config file for NsJail instead of command-line options
* Map `65534` (nobody) user & group inside the user namespace to `65534` outside the namespace rather than mapping to current uid/guid (which was `0` AKA root)

#### Unprivileged Docker container

I've explored running the container as unprivileged, which would make Docker mount neither devfs nor sysfs. However, so far this has been problematic mainly because the NsJail cgroup needs to be created. That requires  `/sys` to be mounted and privileges to write to it. There are some experimental Docker features which allow for privileges while building, but I couldn't manage to create the cgroup during the build. In retrospect, it's probably because `/sys` wasn't mounted anyway (would need to double check if it gets mounted during build).

Regardless, it is questionable what advantages running as unprivileged would have when NsJail is already being employed and now a (hopefully) proper chroot jail is configured. Of course, giving privileges when they aren't needed is always something one would rather not do. Doing away with privileges in this case just doesn't seem trivial to accomplish.

#### Ownership inside NsJail

Inside NsJail, `ls` shows all mounted files and directories are owned by `65534`. Though theoretically everything is mounted as read-only, this is still a bit concerning since the owner has write permissions. 

However, the output of `ls` may be a bit misleading: according to one of the maintainers of NsJail, it is overflowing to `65534` because the root uid/guid is not mapped in the user namespace. If root is mapped to e.g. `12345` inside, then all files show as belonging to `12345` instead, which seems to be correct because everything in the Docker container is owned by root (`USER` is never specified in the Dockerfiles so I guess Docker defaults to using root when copying from the host).

### DIFF
--- a/snekbox/nsjail.py
+++ b/snekbox/nsjail.py
@@ -128,7 +128,7 @@ class NsJail:
                 "--cgroup_pids_mount", str(CGROUP_PIDS_PARENT.parent),
                 "--cgroup_pids_parent", CGROUP_PIDS_PARENT.name,
                 "--",
-                self.python_binary, "-Iq", "-c", code
+                self.python_binary, "-Iqu", "-c", code
             )
 
             msg = "Executing code..."
--- a/snekbox/nsjail.py
+++ b/snekbox/nsjail.py
@@ -35,7 +35,7 @@ class NsJail:
     Default NsJail configuration (snekbox.cfg):
 
     - All mounts are read-only
-    - Time limit of 2 seconds
+    - Time limit of 5 seconds
     - Maximum of 1 PID
     - Maximum memory of 52428800 bytes
     - Loopback interface is down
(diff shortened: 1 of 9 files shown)


snekbox/nsjail.py
```diff
@@ -24,16 +24,17 @@
...
 class NsJail:
     """
     Core Snekbox functionality, providing safe execution of Python code.
 
-    NsJail configuration:
+    Default NsJail configuration (snekbox.cfg):
 
-    - Root directory is mounted as read-only
+    - All mounts are read-only
     - Time limit of 2 seconds
```
SebastiaanZ: If I'm not mistaken, @jos-b changed the time_limit to 5 seconds in his commit.
```suggestion
    - Time limit of 5 seconds
```

PR #59, opened 2020-02-21
Use port mapping, change container name to snekbox.

### LINKED ISSUE #58: Don't run in host mode
Network mode is currently set to "host". That's considered pretty insecure; instead, we should just connect port 8060 to the host. I tried that way and it worked for me, is there a different reason we should be using host mode?

scragly:
Closes #58 

The provided compose currently runs in host mode, which is unnecessary. This adjusts it back to only expose the required port.

The service name has been changed to "snekbox" instead also, as we're moving away from using the `pd` prefix on our services. It also doesn't make too much sense to have such a prefix and shortened name if we're to accommodate public usage.

The hostname field has been removed, and container_name field added. By default, the hostname will be set to the same as the container name unless otherwise specified. To ensure the container name and hostname is predictably the same as the service name, container_name been explicitly specified.


MarkKoz:
The devsh script is failing to execute Python via NsJail. I was running into this issue locally too. Guess it really is a big problem...

PR #64, opened 2020-03-05
Add pep8-naming and more pre-commit hooks

### LINKED ISSUE #63: Add pep8-naming to the linting toolchain.
We've decided to use this plugin in our linting toolchain, which allows us to test things like variable name casing (no camelCase please!) and whether classes and methods use `cls` and `self`.

## Implementation
Here are the errors it can catch:

code | sample message
-- | --
N801 | class names should use CapWords convention
N802 | function name should be lowercase
N803 | argument name should be lowercase
N804 | first argument of a classmethod should be named 'cls'
N805 | first argument of a method should be named 'self'
N806 | variable in function should be lowercase
N807 | function name should not start and end with '__'
  |  
N811 | constant imported as non constant
N812 | lowercase imported as non lowercase
N813 | camelcase imported as lowercase
N814 | camelcase imported as constant
N815 | mixedCase variable in class scope
N816 | mixedCase variable in global scope

We have not decided exactly which of these we should ignore, so it will be up to the person who handles this issue to discuss this with the core devs and come to a consensus on whether we should ignore some of these. It will probably be easier to see which ones here are not necessary after actually running a lint with this plugin on our current codebase.

The person implementing this will also need to fix all the linting errors that occur once this has been added to the toolchain.

## Steps
- [ ] Add the plugin to the toolchain
- [ ] Run it and look through the errors that are created. Are any of them out of touch with our style guidelines, and should we ignore some of these rules? Discuss with the core developers if you think so.
- [ ] Fix all the linting problems.

MarkKoz:
## Relevant Issues

Resolves #63
python-discord/organisation#138

## Description

New hooks were added for pre-commit and they will run in CI too. The `pipenv run lint` script will now run all the new hooks, including flake8.

### Hooks added

A couple of these hooks automatically apply fixes. However, they still report failure and leave any changes they make uncommitted. Therefore, the user has to commit the automatic fixes.

* `check-merge-conflict` - Check for files that contain merge conflict strings.
* `check-toml` - Attempts to load all toml files to verify syntax. 
* `check-yaml` - Attempts to load all yaml files to verify syntax. 
* `end-of-file-fixer` - Makes sure files end in a newline and only a newline.
* `mixed-line-ending` - Replaces mixed line endings with LF. 
* `trailing-whitespace` - Trims trailing whitespace.
* `python-check-blanket-noqa` - Enforce that `noqa` annotations always occur with specific codes

## Reasoning

* Dev dependencies were updated and re-pinned because why not? They were getting old enough. I'm just following along with the other repos.
* No cache used for pre-commit because it runs in a Docker container. I suppose it could be copied to the container from the cache on the host, but that doesn't seem worthwhile.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,9 +1,15 @@
 [flake8]
-max-line-length=100
-application_import_names=snekbox
-docstring-convention=all
-ignore=
-    P102,B311,W503,E226,S311,
+allow-untyped-nested = true
+application_import_names = snekbox
+docstring-convention = all
+import-order-style = pycharm
+inline-quotes = "
+max-line-length = 100
+
+exclude = __pycache__,.cache,user_base,venv,.venv
+
+ignore =
+    W503,
     # Missing Docstrings
     D100,D104,D105,D107,
     # Docstring Whitespace
@@ -14,9 +20,5 @@ ignore=
     D400,D401,D402,D404,D405,D406,D407,D408,D409,D410,D411,D412,D413,D414,D416,D417,
     # Type Annotations
     ANN002,ANN003,ANN101,ANN102,ANN204,ANN206
-exclude=
-    __pycache__,.cache,
-    venv,.venv,user_base
-per-file-ignores=tests/*:D1,ANN
-import-order-style=pycharm
-inline-quotes="
+
+per-file-ignores = tests/*:D1,ANN
--- a/.github/workflows/lint-test-build-push.yaml
+++ b/.github/workflows/lint-test-build-push.yaml
@@ -102,6 +102,12 @@ jobs:
           docker exec snekbox_test /bin/bash -c
           'pipenv install --system --deploy --dev && pip install numpy'
 
+      # Required by pre-commit.
+      - name: Install git
+        run: >-
+          docker exec snekbox_test /bin/bash -c
+          'apt-get -y update && apt-get install -y git=1:2.20.*'
+
       # Skip the flake8 hook because the following step will run it.
       - name: Run pre-commit hooks
         run: >-
--- a/.github/workflows/lint-test-build-push.yaml
+++ b/.github/workflows/lint-test-build-push.yaml
@@ -102,6 +102,12 @@ jobs:
           docker exec snekbox_test /bin/bash -c
           'pipenv install --system --deploy --dev && pip install numpy'
 
+      # Skip the flake8 hook because the following step will run it.
+      - name: Run pre-commit hooks
+        run: >-
+          docker exec snekbox_test /bin/bash -c
+          'SKIP=flake8 pre-commit run --all-files'
+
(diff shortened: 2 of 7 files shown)


MarkKoz:
pre-commit fails to run cause git is not installed in the container. May need to rethink this...

MarkKoz:
Build failing because pre-commit cannot find shellcheck

> Test shell scripts with shellcheck.......................................Failed
> - hook id: shellcheck
> - exit code: 127
> 
> /root/.cache/pre-commit/repo_gx8xowx/pre_commit_hooks/shellcheck: 13: /root/.cache/pre-commit/repo_gx8xowx/pre_commit_hooks/shellcheck: shellcheck: not found
>

MarkKoz:
Now I don't know why it fails. Maybe cause it's an older version (0.5.0). Shellcheck 0.7.1 doesn't complain locally.

```
Test shell scripts with shellcheck.......................................Failed
- hook id: shellcheck
- exit code: 1

In scripts/check_dockerfiles.sh line 22:
    if [[ -v build_cache["${branch}"] ]]; then
    ^-- SC1009: The mentioned syntax error was in this if expression.
       ^-- SC1073: Couldn't parse this test expression. Fix to allow more checks.
             ^-- SC1019: Expected this to be an argument to the unary condition.
                                     ^-- SC1020: You need a space before the ]].
                                     ^-- SC1072: Missing space before ]. Fix any mentioned problems and try again.


In scripts/dev.sh line 17:
    && printf " done!\n" || exit "$?"
                     ^-- SC1117: Backslash is literal in "\n". Prefer explicit escaping: "\\n".


In scripts/dev.sh line 28:
            && printf " done!\n" || exit "$?"
                             ^-- SC1117: Backslash is literal in "\n". Prefer explicit escaping: "\\n".
```

MarkKoz:
Okay, I give up on ShellCheck. I can't install from the tarball cause the container doesn't have `xz` to decompress it. I _could_ install `xz` but this just isn't worth it any more. I should honestly rewrite the shell scripts in Python anyway.

MarkKoz:
@Xithrius Are you planning on finishing this? If I remember correctly, @SebastiaanZ was since he did #79 but I think he's been too busy. If you are, then the CI stuff here still needs to be migrated to GH Actions. Otherwise, I should find time to finish this myself since it is my PR after all.

By the way, why are you putting "rebase" in your commit summaries when it looks like they are merge commits, not rebases? If you were to actually rebase master, that would have required a force push.

Xithrius:
> @Xithrius Are you planning on finishing this?

I don't know how to properly test this PR, so it would probably be best for you to finish this up.

> By the way, why are you putting "rebase" in your commit summaries when they look like they are merge commits, not rebases?

I've accidentally used the wrong terminology, thank you for pointing that out.

jb3 (approved):
This all looks good. Everything works perfect locally.

PR #65, opened 2020-03-08
Add sentry integration

jos-b:
Integrates Sentry with Snekbox.

Adds the FalconIntegration as well to handle Internal Server errors.

Note: This could not be tested locally because Docker & Pipenv on MacOS are hyper-broken for me.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -23,6 +23,7 @@ networkx = "~= 2.4"
 beautifulsoup4 = "~= 4.8"
 pyyaml = "~= 5.2"
 toml = "~= 0.10"
+sentry-sdk = {extras = ["falcon"], version = "0.14.2"}
 
 [dev-packages]
 coverage = ">= 4.4.2, == 4.*"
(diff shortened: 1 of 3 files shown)


Pipfile
```diff
@@ -23,6 +23,7 @@ networkx = "~= 2.4"
 beautifulsoup4 = "~= 4.8"
 pyyaml = "~= 5.2"
 toml = "~= 0.10"
+sentry-sdk = {extras = ["falcon"], version = "0.14.2"}
```
MarkKoz: This looks incorrect, but haven't tested. Either way it's a good idea to pin it a bit more loosely.

```suggestion
sentry-sdk = {extras = ["falcon"], version = "~= 0.14.2"}
```

jos-b:
```
pipenv is broken.

the world is crumbling.
```

MarkKoz:
Will someone else need to lock and push it?

jos-b:
I'm locking it now, I fixed pipenv by deleting all my `.DS_Store` files since pipenv cannot handle it.

PR #70, opened 2020-03-29
add --dev flag to readme

neonsea:
In the "development environment" section of the readme, it instructs to run `pipenv sync` to install dependencies and `pipenv run precommit` immediately after to install the pre-commit hooks. Since the `pre-commit` package (and related packages) is listed in the pipfile under `[dev-packages]`, this command fails.

Solution: instruct the user to install all dependencies with the `--dev` flag.


MarkKoz:
Thanks for catching that

PR #71, opened 2020-03-30
Support for running unix commands

neonsea:
### Description
This is a patchset to utilise snekbox to run Unix commands (namely GNU/Linux Debian stable).

### Implementation
It implements a new API endpoint `/unixcmd`, the usage of which is analogue to the already existing `/eval` endpoint - but instead of evaluating the `input` parameter value as Python code, it is executed in a real Linux environment in a bash shell. As with `/eval`, it returns `output` with the combined output of `stdout` and `stderr` and whatever return code bash returned, which is usually the return code of the last command executed.

Currently, this is implemented using a Debian root filesystem. When building the `venv` Docker image, unless the variable `NO_LINUXFS` is set, `debootstrap` is used to bootstrap a Debian stable image into `/snekbox/linuxfs`. When `/unixcmd` is called, snekbox instructs NsJail to bind this directory as `/` (as read-only) and run its `/bin/bash`. 

It should be noted this environment is technically and practically different than the one snekbox evaluates Python code in. The main differences are that the maximum PID (process) count is increased to 10, which is a design requirement (bash would be unable to run anything otherwise), and that this environment has a _full_ filesystem hierarchy as opposed to just the files that snekbox needs to run Python. As such, I feel comfortable only letting a higher rank such as Helpers+ run this command, but that is a bot interface implementation detail rather than snekbox's.

### Rationale
I've noticed that there are a lot of times in the `#unix` channel when someone asks for help with something which could easily be demonstrated on a "real" Unix system, but harder to explain as a concept. This feature would be useful in such cases.

### Example usage
Request:
```http
POST /unixcmd HTTP/1.1
Host: localhost:8060
Content-Type: application/json
Content-Length: 24

{"input":"id; uname -a"}
```

Response:
```http
HTTP/1.1 200 OK
Server: gunicorn/19.10.0
Date: Mon, 30 Mar 2020 16:01:11 GMT
Connection: close
content-type: application/json
content-length: 185

{"stdout": "uid=65534(nobody) gid=65534(nogroup) groups=65534(nogroup)\nLinux snekbox 5.5.11-arch1-1 #1 SMP PREEMPT Sun, 22 Mar 2020 16:33:15 +0000 x86_64 GNU/Linux\n", "returncode": 0}
```

### To-Do
- [x] fix broken test `test_subprocess_resource_unavailable` (do not pass PID count as nsjail argument)
- [x] write tests for new code
- [ ] Bash prints "Resource temporarily unavailable" three times before exiting when NsJail limits are hit (time limit, pid limit). Need to find a way to display this in a user-friendly manner
- [x] write logic for when `/unixcmd` is called when no linuxfs is set up. Currently, NsJail returns blank output and return code -1 (255)
- [ ] create PR for bot interface implementation
- [ ] `UnixCmdResource` and `EvalResource` are the same apart from the nsjail function being called. Find a way to remove duplication
- [x] fix logparser logic
- [ ] update documentation

Comments, ideas, criticism, etc all very welcome.

### DIFF
--- a/config/unixcmd.cfg
+++ b/config/unixcmd.cfg
@@ -1,0 +1,62 @@
+name: "unixcmd"
+description: "Execute Unix shell commands"
+
+mode: ONCE
+hostname: "snekbox"
+cwd: "/"
+
+time_limit: 6
+
+keep_env: false
+envar: "LANG=en_US.UTF-8"
+envar: "OMP_NUM_THREADS=1"
+envar: "OPENBLAS_NUM_THREADS=1"
+envar: "MKL_NUM_THREADS=1"
+envar: "VECLIB_MAXIMUM_THREADS=1"
+envar: "NUMEXPR_NUM_THREADS=1"
+
+keep_caps: false
+
+rlimit_as: 700
+
+clone_newnet: true
+clone_newuser: true
+clone_newns: true
+clone_newpid: true
+clone_newipc: true
+clone_newuts: true
+clone_newcgroup: true
+
+uidmap {
+    inside_id: "65534"
+    outside_id: "65534"
+}
+
+gidmap {
+    inside_id: "65534"
+    outside_id: "65534"
+}
+
+mount_proc: false
+
+mount {
+    src: "/snekbox/linuxfs"
+    dst: "/"
+    is_bind: true
+    rw: false
+}
+
+cgroup_mem_max: 52428800
+cgroup_mem_mount: "/sys/fs/cgroup/memory"
+cgroup_mem_parent: "NSJAIL"
+
+cgroup_pids_max: 10
+cgroup_pids_mount: "/sys/fs/cgroup/pids"
+cgroup_pids_parent: "NSJAIL"
+
+iface_no_lo: true
+
+exec_bin {
+    path: "/bin/bash"
+    arg: "-c"
+}
--- a/config/unixcmd.cfg
+++ b/config/unixcmd.cfg
@@ -9,11 +9,6 @@ time_limit: 6
 
 keep_env: false
 envar: "LANG=en_US.UTF-8"
-envar: "OMP_NUM_THREADS=1"
-envar: "OPENBLAS_NUM_THREADS=1"
-envar: "MKL_NUM_THREADS=1"
-envar: "VECLIB_MAXIMUM_THREADS=1"
-envar: "NUMEXPR_NUM_THREADS=1"
 
 keep_caps: false
 
--- a/docker/venv.Dockerfile
+++ b/docker/venv.Dockerfile
@@ -1,6 +1,7 @@
 FROM pythondiscord/snekbox-base:latest
 
 ARG DEV
+ARG NO_LINUXFS
 ENV PIP_NO_CACHE_DIR=false \
     PIPENV_DONT_USE_PYENV=1 \
     PIPENV_HIDE_EMOJIS=1 \
@@ -15,6 +16,8 @@ RUN if [ -n "${DEV}" ]; \
     else \
         pipenv install --deploy --system; \
     fi
+
+RUN [ -z "${NO_LINUXFS}" ] && pipenv run bootstrap-linux
 
 # At the end to avoid re-installing dependencies when only a config changes.
 COPY config/ /snekbox/config
--- a/docker/venv.Dockerfile
+++ b/docker/venv.Dockerfile
@@ -17,7 +17,11 @@ RUN if [ -n "${DEV}" ]; \
         pipenv install --deploy --system; \
     fi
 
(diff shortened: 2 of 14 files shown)


Akarys42:
I looks like a pretty good and very clean implementation. Although, do we want snekbox to support both python and unix commands, or create another project for it? The name itself doesn't really make sense anymore imo

neonsea:
I think the name makes sense - the project is still a Python implementation :)

On a more serious note, if this were a separate project, it would have a lot of overlapping code with snekbox as it would technically be largely the same backend. Maintaining two different projects for what are essentially the same things doesn't seem advantageous to me, but I want to hear more thoughts on this.

docker/venv.Dockerfile
```diff
@@ -16,5 +17,7 @@ RUN if [ -n "${DEV}" ]; \
         pipenv install --deploy --system; \
     fi
 
+RUN [ -z "${NO_LINUXFS}" ] && pipenv run bootstrap-linux
```
MarkKoz: Avoid using `pipenv run` inside Docker because it will create a venv. It cannot know that a `--system` install was done.

MarkKoz:
Yes, the name isn't totally accurate but whatever - it's just a name. For simplicity's sake, it makes most sense to keep everything together. You're correct about the similarities they would share. That would either involve duplicating code (hard to maintain) or creating a third repo for a common library (adds complexity to the dev process).

MarkKoz:
Another thing you'll need to do is update documentation such as README.md and docstrings.

docker/venv.Dockerfile
neonsea: Any suggestions for this?

MarkKoz: Invoke the command directly in the Dockerfile. What is the rationale for having it in the Pipfile in the first place?

neonsea: I originally had the linuxfs on the host filesystem - having it in the Pipfile meant I could rapidly redeploy just the linuxfs without having to touch anything else. You're right though, I've reworked the logic and there's no reason for it to be there anymore. I'll move it straight into the Dockerfile.

tests/nsjail/test_unix.py
```diff
@@ -0,0 +1,82 @@
...
+        self.assertEqual(result.stdout, "mount: failed to read mtab: No such file or directory\n")
+        self.assertEqual(result.stderr, None)
+
+    def test_readonly_filesystem(self):
+        cmd = dedent("""
+            for location in "/ /dev /tmp /proc /bin"; do
+                touch $location/test 2>/dev/null
+                [ $? -eq 0 ] && echo Success
+            done
+        """)
```
MarkKoz, line 67: I think these should be subtests like the Python 3 test. It'll give a better output upon failure (like which directory failed).

tests/nsjail/test_unix.py
```diff
@@ -80,3 +80,17 @@ def test_cannot_remount_rootfs(self):
...
+        cmd = dedent("""
+            whoami
+        """)
+
+        result = self.nsjail.unix(cmd)
+        self.assertEqual(result.returncode, None)
+        self.assertEqual(result.stdout, "LinuxFS not set up")
+        self.assertEqual(result.stderr, None)
+
+        self.nsjail.shell_binary = orig_binary  # clean up
```
MarkKoz, line 96: Clean up shouldn't be necessary since the `setUp` fixture is executed separately for every test. In other words, the nsjail instance is re-created for each test rather than shared among them.

snekbox/api/resources/unixcmd.py
```diff
@@ -44,7 +44,7 @@ def on_post(self, req: falcon.Request, resp: falcon.Response) -> None:
         Some noteworthy exceptions:
 
         - None
-            The NsJail process failed to launch
+            The NsJail process failed to launch. This will happen if LinuxFS is not set up.
         - 137 (SIGKILL)
             Typically because NsJail killed the Python process due to time or memory constraints
```
MarkKoz, line 49: "Python process" is not accurate for this.

config/unixcmd.cfg
```diff
@@ -0,0 +1,62 @@
...
+
+time_limit: 6
+
+keep_env: false
+envar: "LANG=en_US.UTF-8"
+envar: "OMP_NUM_THREADS=1"
+envar: "OPENBLAS_NUM_THREADS=1"
+envar: "MKL_NUM_THREADS=1"
+envar: "VECLIB_MAXIMUM_THREADS=1"
+envar: "NUMEXPR_NUM_THREADS=1"
```
MarkKoz, line 16: These were for numpy so they aren't necessary here.

neonsea:
Thanks for the comments, I'll get on them.

neonsea:
Closed after discussion about the risk vs reward aspect in #dev-contrib.

PR #79, opened 2020-11-18
Migrate to GitHub Actions and GitHub Container Registry

SebastiaanZ:
I've migrated the build pipeline to GitHub Actions and changed the container registry to GitHub Container Registry. In the process, I've also made some changes to our docker setup and caching.

- **Note:** This PR does not yet take #64 into account. I will update that PR later.

- **Note II:** We need to get to a mergeable state sooner rather than later, as this is currently blocking our migration to the Kubernetes cluster.

### We are now using a single multi-stage Dockerfile

Instead of three separate dockerfiles, we are now using a single multi-stage Dockerfile that can be used to build the three images
we want using build targets.

In part, this is because we're now using the [docker buildx build action](https://github.com/docker/build-push-action) currently recommended by docker. This new engine runs in a sandboxed mode, meaning that while it can export built images to `docker` running in the host, it cannot import local images from it to base builds on. Building truly separate images, from different dockerfiles, would mean we'd have to push and pull the intermediate images.

By having a single multi-stage Dockerfile, we can still build the intermediate containers and cache their layers, but we don't have to reference other image tags in the Dockerfile. 

This new builder back-end comes with some interesting additional options in terms of caching that should work similar to the custom caching script @MarkKoz wrote for Azure (see below).

### Docker builds are now cached within GitHub Actions

The builds are now cached using the GitHub Actions cache of the build cache directory. The cache keys try to match a cache generated by a build that matches the current build as closely as possible. In case of a cache miss, we fall back to caching from the latest image pushed to the container repository.

### The `base` and `venv` images now have an inline cache manifest

In order to fall back intelligently to caching from the repository, the final build and push action for the `base` and `venv` images now includes an "inline" cache manifest with the image. This means that the build process can inspect, without pulling the entire image, if it makes sense to pull layers to speed up the build.

The other options, pushing a cache manifest separately (not inline), is currently not supported by GHCR.

The custom caching script has been removed.

### Linting errors are now added as GitHub Actions annotations

Just like for some of our other pipelines, linting now generates annotations if linting errors are observed.

### Coverage is pushed to coveralls.io

A coverage summary is now pushed to coveralls.io. Each CI run will get a unique job that's linked in the CI output. If the run is attached to a PR, coveralls.io will automatically add a check link with the coverage result to the PR as well.

### The README.md, Pipfile, docker-compose, and scripts have been updated

As we now need to pull from and link to the GHCR, I've updated the other files to reflect these changes, including Pipfile run commands. I've also changed the CI badge and added a coveralls.io badge.


.github/workflows/lint-test-build-push.yaml
```diff
@@ -0,0 +1,207 @@
...
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+
+jobs:
+  lint-test:
```
SebastiaanZ, line 11: It would probably be better to update this name to reflect the entire scope of the job:

```suggestion
  lint-test-build-push:
```

.github/workflows/lint-test-build-push.yaml
```diff
@@ -0,0 +1,207 @@
...
+
+      - name: Start Container
+        run: "docker run \
+          --tty \
+          --detach \
+          --name snekbox_test \
+          --privileged \
+          --hostname pdsnk-dev \
+          -e PYTHONDONTWRITEBYTECODE=1 \
+          -e PIPENV_PIPFILE='/snekbox/Pipfile\' \
```
SebastiaanZ, line 89: What's up with this weird escape? And why does the pipeline work with it?

```suggestion
          -e PIPENV_PIPFILE='/snekbox/Pipfile' \
```

jb3 (approved):

.github/workflows/lint-test-build-push.yaml
```diff
@@ -0,0 +1,207 @@
...
+
+
+jobs:
+  lint-test-build-push:
+    runs-on: ubuntu-latest
+    env:
+      # Determine whether or not we should build the
+      # final production image and push it to GHCR.
+      production_build: ${{ github.event_name != 'pull_request' &&
+        github.ref == 'refs/heads/sebastiaan/backend/cache-docker-images' }}
```
MarkKoz, line 17: Surely this ref needs to be changed.

.github/workflows/lint-test-build-push.yaml
```diff
@@ -0,0 +1,207 @@
...
+
+    steps:
+      # Create a short SHA-tag to tag built images
+      - name: Create SHA Container Tag
+        id: sha_tag
+        run: |
+          tag=$(cut -c 1-7 <<< $GITHUB_SHA)
+          echo "::set-output name=tag::$tag"
+      - name: Checkout code
```
MarkKoz, line 26: It's more readable with a space between the steps.
```suggestion

      - name: Checkout code
```

.github/workflows/lint-test-build-push.yaml
```diff
@@ -0,0 +1,207 @@
...
+      - name: Login to Github Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GHCR_TOKEN  }}
+
+      # Set up a caching directory for image layers. According to the docker
+      # documentation, it's recommended to use a SHA-based key to get the
+      # greatest change of finding the most relevant cached layer. We fall
```
MarkKoz, line 41:
```suggestion
      # greatest chance of finding the most relevant cached layer. We fall
```

.github/workflows/lint-test-build-push.yaml
```diff
@@ -0,0 +1,207 @@
...
+        uses: docker/setup-buildx-action@v1
+
+      - name: Login to Github Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GHCR_TOKEN  }}
+
+      # Set up a caching directory for image layers. According to the docker
```
MarkKoz, line 39:
```suggestion
      # Set up a caching directory for image layers. According to the Docker
```

.github/workflows/lint-test-build-push.yaml
```diff
@@ -0,0 +1,207 @@
...
+      # down to more generic containers by then matching by GitHub branch,
+      # to use cache generated earlier in the same branch, and finally to
+      # the latest cache in general. The `v0` is purely a cache version
+      # indicator that can be incremented manually if we want to invalidate
+      # old caches completely.
+      - name: Cache Image Layers
+        uses: actions/cache@v2
+        with:
+          path: /tmp/.buildx-cache
+          key: ${{ runner.os }}-v0-buildx-${{ github.ref }}-${{ github.sha }}
```
MarkKoz, line 51: Will `github.sha` only have a cache hit for a cache generated by a previous run for the same commit? This would only happen for manual re-runs presumably, right? Therefore, is it correct that it's mostly relying on restore keys to get a cache hit?

.github/workflows/lint-test-build-push.yaml
```diff
@@ -0,0 +1,207 @@
...
+          restore-keys: |
+            ${{ runner.os }}-v0-buildx-${{ github.ref }}-
+            ${{ runner.os }}-v0-buildx-
```
MarkKoz, line 54: If it always restores a cache, doesn't this mean the cache will get bloated with old layers as time goes on?

.github/workflows/lint-test-build-push.yaml
```diff
@@ -0,0 +1,207 @@
...
+
+      # Build the image we need for testing/linting the current codebase,
+      # without pushing the image to the GHCR. Instead, we load it into
+      # the runner's docker environment so we can run it later. The
```
MarkKoz, line 58:
```suggestion
      # the runner's Docker environment so we can run it later. The
```

.github/workflows/lint-test-build-push.yaml
```diff
@@ -0,0 +1,207 @@
...
+          --name snekbox_test \
+          --privileged \
+          --hostname pdsnk-dev \
+          -e PYTHONDONTWRITEBYTECODE=1 \
+          -e PIPENV_PIPFILE='/snekbox/Pipfile' \
+          -e ENV=\"${PWD}/scripts/.profile\" \
+          --volume \"${PWD}\":\"${PWD}\" \
+          --workdir \"${PWD}\" \
+          --entrypoint /bin/bash \
+          ghcr.io/python-discord/snekbox-venv:${{ steps.sha_tag.outputs.tag }}"
```
MarkKoz, line 94: It may be possible to make this nicer by using the `>` YAML block scalar (or `>-` if a single newline at the end would cause problems), which replaces newlines with spaces. I believe this would get rid of the surrounding double quotes and `\` at the end of every line.

Pipfile
```diff
@@ -61,26 +60,29 @@ snekbox = "gunicorn -c config/gunicorn.conf.py snekbox.api.app"
...
 builddev = """
     docker build \
-        -t pythondiscord/snekbox-venv:dev \
-        -f docker/venv.Dockerfile \
+        -t ghcr.io/python-discord/snekbox-venv:dev \
+        -f docker/Dockerfile \
+        --target venv \
         --build-arg DEV=1 \
         .
 """
```
MarkKoz, line 88: I don't think the additional scripts still have a use given that it's a single image now.

MarkKoz (changes requested):
Since there's only one Dockerfile now, it's redundant to keep it within the `Docker` subdirectory.

Pipfile
MarkKoz: `builddev` may be useful, but probably not since `dev.sh --build` and `testb` exist.

.github/workflows/lint-test-build-push.yaml
SebastiaanZ: That's embarrassing. Good catch. I've been staring at this for far too long.

.github/workflows/lint-test-build-push.yaml
SebastiaanZ: I was wondering the same thing, but I'm following a [recipe documented in the README for the docker action](https://github.com/docker/build-push-action#leverage-github-cache). I've tried looking for more information on this, but I haven't found it yet.

.github/workflows/lint-test-build-push.yaml
SebastiaanZ: I was wondering the same thing, but I'm following a [recipe documented in the README for the docker action](https://github.com/docker/build-push-action#leverage-github-cache). I've tried looking for more information on this, but I haven't found it yet.

MarkKoz: I think what is restored from the partial cache hit is written to the new cache created when the workflow ends, since the cache action just saves whatever is in a particular directory. Maybe running `docker system prune -f` before ending the workflow will avoid this problem. Hopefully that will not clear stuff related to the images that were just built. However, I think the `-a` option might (how would it know which images are "in use").

SebastiaanZ: I'm not entirely sure if that will work for a `type=local` cache, going by [this issue](https://github.com/docker/buildx/issues/310). Currently, I'm thinking of running this with a repository-only cache to see how that affects build speeds. If the difference in build speed is negligible, maybe we can just do away with local caching entire and rely on GHCR for our layer cache.

SebastiaanZ: That seems to work well. I think it may lead to faster builds in some situations and longer builds in some others, where the `latest` image is a bit out of sync with cache-invalidating changes made in a feature branch. However, if that's not the cache, this seems to be faster than using the local cache option.

A fully non-cached build took 3m7s; a fully cached build took 1m45s, using just the repository cache. I think we should just be happy with that and not worry about the local cache that could potentially get bloated.

If we notice systematically longer builds in the future, we could try another approach (e.g, making the local cache branch specific and manually creating a clean cache dir once per x days for master).

MarkKoz: One thing I forgot is that when it restores a cache from master, the new cache it writes will be scoped locally rather to master. Therefore, PRs will only bloat caches for PRs. Since PRs don't tend to have very long lifetimes, bloating shouldn't be as bad. It's still not ideal, since every new commit a PR makes will add new layers to the cache without getting rid of layers specific to previous commits. However, I imagine the layers will be small (basically just the source code copy step) unless every commit is constantly changing the venv or base layers.

I think the local cache should be disabled on master because the bloat could live on indefinitely, assuming enough activity that the cache doesn't expire. A local cache isn't generally useful on master for two reasons:

1. I don't believe the final merge to master has access to the PR's cache scope. 
2. If the previous run on master succeeded, then the images will be available from the registry, making the cache redundant.

.github/workflows/lint-test-build-push.yaml
SebastiaanZ: This was included, but not recognized as such by GH, in one of the commits I pushed.

.github/workflows/lint-test-build-push.yaml
SebastiaanZ: It is, yes. I took this approach from an example in the README.md of Docker's build-push action.

MarkKoz (changes requested):

Dockerfile
```diff
@@ -32,3 +32,32 @@ RUN pip install pipenv==2020.11.4
...
+
+# At the end to avoid re-installing dependencies when only a config changes.
+COPY config/ /snekbox/config
+
+FROM venv
+
+ENTRYPOINT ["gunicorn"]
+CMD ["-c", "config/gunicorn.conf.py", "snekbox.api.app"]
+
+COPY docker /snekbox
```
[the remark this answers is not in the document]
SebastiaanZ: This shouldn't have happened. I'm not sure why it did either. I'll change it back.

MarkKoz (approved):

PR #80, opened 2020-11-19
Kubernetes auto-deployment

lemonsaurus:
This PR adds two things:
- A k8s deployment manifest
- A deploy instruction inside the build stage of the CI.


MarkKoz (commented):
I believe both k8s steps need the `if: env.production_build == 'true'` condition.

lemonsaurus:
Right you are, thanks @MarkKoz!

jb3 (approved):

PR #81, opened 2020-11-20
Limit STDOUT to prevent OOM events in container

SebastiaanZ:
Recently, we discovered that for some code inputs, snekbox would get into an OOM event on the container level, seemingly bypassing the memory restrictions laid on code execution by NSJail.

After investigating the issue, we identified the culprit to be the STDOUT pipe we use to get output back from NSJail: As output is piped out of the jailed process, it will be collected outside of the NSJail in the main container process instead. This meant that our initial attempts of limiting the allowed filesize within the NSJail failed, as the OOM happened outside of the jailed environment.

To mitigate the issue, I've written a loop that consumes the STDOUT pipe in chunks of 100 characters. Once the size of the accrued output reaches a certain limit (currently set to 1 MB), we send a SIGTERM signal to NSJail to terminate itself. The output up to that point will be relayed back to the caller.

A minimal code snippet to trigger the event and the mitigation:

```py
while True:
    print("1234567890")
```

I've included a test for this vulnerability in [`tests/test_nsjail.py`](https://github.com/python-discord/snekbox/blob/5639a400e633a48c8564a61a035485933229d4f5/tests/test_nsjail.py#L178).

### DIFF
--- a/snekbox/nsjail.py
+++ b/snekbox/nsjail.py
@@ -104,6 +104,42 @@ class NsJail:
                 # Treat fatal as error.
                 log.error(msg)
 
+    @staticmethod
+    def _consume_stdout(nsjail: subprocess.Popen) -> str:
+        """
+        Consume STDOUT, stopping when the output limit is reached or NsJail has exited.
+
+        The aim of this function is to limit the size of the output received from
+        NsJail to prevent container from claiming too much memory. If the output
+        received from STDOUT goes over the OUTPUT_MAX limit, the NsJail subprocess
+        is asked to terminate with a SIGTERM.
+
+        Once the subprocess has exited, either naturally or because it was terminated,
+        the output up to that point is returned as a string.
+        """
+        output_size = 0
+        output = []
+
+        # We'll consume STDOUT as long as the NsJail subprocess is running
+        while nsjail.poll() is None:
+            # Read 100 characters from the STDOUT stream
+            chars = nsjail.stdout.read(100)
+            chars_size = sys.getsizeof(chars)
+
+            # Check if these characters take us over the output limit
+            if output_size + chars_size > OUTPUT_MAX:
+                # Ask nsjail to terminate itself using SIGTERM
+                nsjail.terminate()
+                break
+
+            output_size += chars_size
+            output.append(chars)
+
+        # Ensure that we wait for the nsjail process to terminate
+        nsjail.wait()
+
+        return "".join(output)
+
     def python3(self, code: str) -> CompletedProcess:
         """Execute Python 3 code in an isolated environment and return the completed process."""
         with NamedTemporaryFile() as nsj_log:
@@ -136,33 +172,14 @@ class NsJail:
             except ValueError:
                 return CompletedProcess(args, None, "ValueError: embedded null byte", None)
 
-            output_size = 0
-            output = []
-
-            # We'll consume STDOUT as long as the subprocess is running
-            while nsjail.poll() is None:
-                # Read 100 characters from the STDOUT stream
-                chars = nsjail.stdout.read(100)
-                chars_size = sys.getsizeof(chars)
-
-                # Check if these characters take us over the output limit
-                if output_size + chars_size > OUTPUT_MAX:
-                    # Ask nsjail to terminate itself using SIGTERM
-                    nsjail.terminate()
-                    break
-
-                output_size += chars_size
-                output.append(chars)
-
-            # Ensure that we wait for the nsjail process to terminate
-            nsjail.wait()
+            output = self._consume_stdout(nsjail)
 
             log_lines = nsj_log.read().decode("utf-8").splitlines()
             if not log_lines and nsjail.returncode == 255:
                 # NsJail probably failed to parse arguments so log output will still be in stdout
-                log_lines = "".join(output).splitlines()
+                log_lines = output.splitlines()
 
             self._parse_log(log_lines)
 
         log.info(f"nsjail return code: {nsjail.returncode}")
-        return CompletedProcess(args, nsjail.returncode, "".join(output), None)
+        return CompletedProcess(args, nsjail.returncode, output, None)
--- a/snekbox/nsjail.py
+++ b/snekbox/nsjail.py
@@ -122,7 +122,7 @@ class NsJail:
 
         # We'll consume STDOUT as long as the NsJail subprocess is running.
         while nsjail.poll() is None:
-            chars = nsjail.stdout.read(100)
+            chars = nsjail.stdout.read(10_000)
             output_size += sys.getsizeof(chars)
             output.append(chars)
 
--- a/snekbox/nsjail.py
+++ b/snekbox/nsjail.py
@@ -27,6 +27,9 @@ NSJAIL_PATH = os.getenv("NSJAIL_PATH", "/usr/sbin/nsjail")
 NSJAIL_CFG = os.getenv("NSJAIL_CFG", "./config/snekbox.cfg")
 MEM_MAX = 52428800
 
+# Limit of stdout bytes we consume before terminating nsjail
+OUTPUT_MAX = 1_000_000  # 1 MB
+
 
 class NsJail:
     """
@@ -124,7 +127,7 @@ class NsJail:
             log.info(msg)
 
             try:
-                result = subprocess.run(
+                nsjail = subprocess.Popen(
                     args,
                     stdout=subprocess.PIPE,
                     stderr=subprocess.STDOUT,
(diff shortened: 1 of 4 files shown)


jerbob (approved):

snekbox/nsjail.py
```diff
@@ -133,11 +136,33 @@ def python3(self, code: str) -> CompletedProcess:
...
+            # We'll consume STDOUT as long as the subprocess is running
+            while nsjail.poll() is None:
+                # Read 100 characters from the STDOUT stream
+                chars = nsjail.stdout.read(100)
+                chars_size = sys.getsizeof(chars)
+
+                # Check if these characters take us over the output limit
+                if output_size + chars_size > OUTPUT_MAX:
+                    # Ask nsjail to terminate itself using SIGTERM
+                    nsjail.terminate()
```
MarkKoz, line 151: SIGKILL is better since it'll result in a consistent error on the front end. Furthermore, SIGKILL cannot be handled by the child process, so it's more robust in this case.

snekbox/nsjail.py
```diff
@@ -133,11 +136,33 @@ def python3(self, code: str) -> CompletedProcess:
             except ValueError:
                 return CompletedProcess(args, None, "ValueError: embedded null byte", None)
 
+            output_size = 0
+            output = []
+
+            # We'll consume STDOUT as long as the subprocess is running
+            while nsjail.poll() is None:
+                # Read 100 characters from the STDOUT stream
```
MarkKoz, line 144: It is not accurate to call them "characters" since UTF-8 allows for a single character to be up to 4 bytes long.

snekbox/nsjail.py
```diff
@@ -133,11 +136,33 @@ def python3(self, code: str) -> CompletedProcess:
...
+                chars = nsjail.stdout.read(100)
```
MarkKoz, line 145: 100 seems really low. I suggest something in the 10's of kilobytes. Since an overflow of that magnitude is still insignificant, it should not discard the overflowed bytes it reads.

snekbox/nsjail.py
```diff
@@ -133,11 +136,33 @@ def python3(self, code: str) -> CompletedProcess:
...
+                chars_size = sys.getsizeof(chars)
+
+                # Check if these characters take us over the output limit
+                if output_size + chars_size > OUTPUT_MAX:
+                    # Ask nsjail to terminate itself using SIGTERM
+                    nsjail.terminate()
+                    break
+
+                output_size += chars_size
+                output.append(chars)
```
MarkKoz, line 155: It'd help readability to move this to a separate function that returns the output.

MarkKoz (changes requested):

snekbox/nsjail.py
MarkKoz: Consider using `io.DEFAULT_BUFFER_SIZE` or setting a custom `bufsize` for `Popen`.

snekbox/nsjail.py
SebastiaanZ: This is a stream in text mode; the size parameter refers to the number of characters in that case, not the number of bytes.

snekbox/nsjail.py
SebastiaanZ: I can increase the number of characters read from the stream, although this increases the uncertainty about the size in bytes we're reading due to UTF-8 characters being up to 4 bytes long.

SebastiaanZ: > Consider using `io.DEFAULT_BUFFER_SIZE` or setting a custom `bufsize` for `Popen`.

This is the default value for Popen due to the negative default value for bufsize (`-1`):

> negative bufsize (the default) means the system default of io.DEFAULT_BUFFER_SIZE will be used.

MarkKoz: > this increases the uncertainty about the size in bytes we're reading due to UTF-8 characters being up to 4 bytes long.

Well, even being 40 KB over is not a big deal.

> This is the default value for Popen due to the negative default value for bufsize

That was worded poorly. I mean that when you read from the stream, use `read(io.DEFAULT_BUFFER_SIZE)` or set the `bufsize` to the same number you use to read from the stream. Not sure if it really makes a difference.

MarkKoz:
Just to double check, there will always be a front-end indication of truncation since 1 MB is 250,000 - 1 million characters, but pastebin has a limit of 100,000 characters or 0.1 - 0.4 MB. Is this correct?

SebastiaanZ:
> Just to double check, there will always be a front-end indication of truncation since 1 MB is 250,000 - 1 million characters, but pastebin has a limit of 100,000 characters or 0.1 - 0.4 MB. Is this correct?

That is correct. My idea was that I did not want to go lower than the pastebin limit, so I used a safe margin. Now I'm thinking that we could also count the number of characters we consume and limit that to, say, 99,900. That means we can always upload the results. We could even include a message like "`\n[output too long: truncated]\n"` at the bottom. What do you think?

MarkKoz:
I think it's fine as-is, though it should be documented that it truncates at around 1 megabyte.

tests/test_nsjail.py
```diff
@@ -174,3 +174,12 @@ def test_output_order(self):
...
         self.assertEqual(result.stderr, None)
+
+    def test_stdout_flood_results_in_graceful_sigterm(self):
+        stdout_flood = dedent("""
+            while True:
+                print('abcdefghij')
+        """).strip()
+
+        result = self.nsjail.python3(stdout_flood)
+        self.assertEqual(result.returncode, -9)
```
MarkKoz, line 185: Is -9 equivalent to 137? Is this some sort of two's complement thing around 127 or 1 byte?

MarkKoz (changes requested):
This looks good. I'd just like a test for the output truncation. The test you added only tests for whether the process is killed.

tests/test_nsjail.py
SebastiaanZ: No, it's what you get back when you terminate a subprocess with a signal in Popen:  [the signal code, but negative](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.returncode). So, that's what we're getting here as well, a `-9`, because SIGKILL is `9`.

MarkKoz: Then this needs to be adjusted somewhere so that the bot will understand that it's still `SIGKILL`, because right now I believe it will only understand `137`.

SebastiaanZ: We could return a manual return code from the consume function:
- If we exceed the limit, we return a manual 137
- If we don't, we propagate the actual exit code.

MarkKoz: How about something more generic to normalise it:

```py
if result.returncode < 0:
    result.returncode = abs(result.returncode) + 128
```

MarkKoz: I guess `-` can be used instead of `abs()`...

snekbox/nsjail.py
```diff
@@ -128,13 +128,19 @@ def _consume_stdout(nsjail: subprocess.Popen) -> str:
...
                 break
 
         # Ensure that we wait for the NsJail subprocess to terminate.
         nsjail.wait()
 
-        return "".join(output)
+        # When you send signal `N` to a subprocess to terminate it using Popen, it
+        # will return `-N` as its exit code. As we normally get `N + 128` back, we
+        # convert negative exit codes to the `N + 128` form.
+        returncode = -nsjail.returncode + 128 if nsjail.returncode < 0 else nsjail.returncode
```
MarkKoz, line 142: Why here and not in the main function? I'd rather avoid returning a tuple.

MarkKoz (changes requested):

snekbox/nsjail.py
SebastiaanZ: > I'd rather avoid returning a tuple.

I'll change it, but what is your motivation for disliking it? Just personal preference?

MarkKoz: I tend to associate it with functions that are broader in scope than they should be. It could also be a sign of clumsiness, like when returning some sort of boolean in addition to data. It may indicate a function can be further refactored. It's also more difficult to document, especially when not using named tuples. Returning two ostensibly unrelated things in a tuple makes the purpose of the function less clear.

In fact, ideally this function would only be concerned with the output and not do anything to the process, but killing the process within doesn't seem avoidable. This also adds to the trouble of clearly documenting it: this function mainly returns the output, but, by the way, it may also kill the process. That may be an important distinction which wouldn't be clear at a glance.

snekbox/nsjail.py
```diff
@@ -116,7 +116,7 @@ def _consume_stdout(nsjail: subprocess.Popen) -> Tuple[int, str]:
         is asked to terminate with a SIGKILL.
 
         Once the subprocess has exited, either naturally or because it was terminated,
-        we return the exit code as an int and the output as a single string.
+        we rturn the output as a single string.
```
MarkKoz, line 119: I don't know where the suggestion button went, but there is a typo here.

MarkKoz (changes requested):

snekbox/nsjail.py
```diff
@@ -101,6 +105,39 @@ def _parse_log(log_lines: Iterable[str]) -> None:
...
+        """
+        Consume STDOUT, stopping when the output limit is reached or NsJail has exited.
+
+        The aim of this function is to limit the size of the output received from
+        NsJail to prevent container from claiming too much memory. If the output
+        received from STDOUT goes over the OUTPUT_MAX limit, the NsJail subprocess
+        is asked to terminate with a SIGKILL.
+
+        Once the subprocess has exited, either naturally or because it was terminated,
+        we rturn the output as a single string.
```
SebastiaanZ, line 119:
```suggestion
        we return the output as a single string.
```

MarkKoz (approved):

PR #85, opened 2020-12-20
Implement Sentry releases

ks129:
Added injecting Git SHA to container using build arguments. On SDK initialization this uses this SHA for release defining. Also added workflow for creating releases.


MarkKoz (commented):
What are the consequences of making a sentry release for a SHA that doesn't go into production (e.g. cause a build failed)? Does it not matter?

ks129:
@MarkKoz So much how I understand by reading docs, this doesn't affect anything. This will be just empty release then.

MarkKoz (approved):

PR #87, opened 2021-01-09
Isolate exposed packages from the project's dependencies

MarkKoz:
This PR accomplishes two things:

1. Prevents snekbox's dependencies from being exposed inside the snekbox Python interpreter.
2. No longer installs pydis-specific third-party packages in the image. This means users will have a blank image to work with and can install the packages they need with the newly added instructions.

Pydis dependencies are now defined in `deployment.yaml` and get installed into a directory which is a mounted Docker volume for persistence.

The implementation uses `PYTHONUSERSITE` to allow `pip install --user` to work to install the packages simply. The snekbox Python interpreter will not use the user site normally. Instead, it will be added to `PYTHONPATH`. The reason it doesn't also just use `PYTHONUSERSITE` is because the [`-S`](https://docs.python.org/3/using/cmdline.html#id3) option is needed to not expose snekbox's dependencies. As a consequence, `-S` also prevents the user site from being added to `sys.path`

### DIFF
--- a/deployment.yaml
+++ b/deployment.yaml
@@ -20,6 +20,9 @@ spec:
             - containerPort: 8060
           securityContext:
             privileged: true
+          volumeMounts:
+            - name: snekbox-user-base-volume
+              mountPath: /snekbox/user_base
           lifecycle:
             postStart:
               exec:
@@ -46,3 +49,7 @@ spec:
                     sympy~=1.6
                     toml~=0.10
                     yarl~=1.6
+      volumes:
+        - name: snekbox-user-base-volume
+          configMap:
+            name: snekbox-user-base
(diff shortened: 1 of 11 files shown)


deployment.yaml
```diff
@@ -20,3 +20,36 @@ spec:
...
+                    python-levenshtein~=0.12.0
+                    pyyaml~= 5.3
+                    scipy~=1.5
+                    sympy~=1.6
+                    toml~=0.10
+                    yarl~=1.6
+      volumes:
+        - name: snekbox-user-base-volume
+          configMap:
+            name: snekbox-user-base
```
jb3, line 55: Swapping out for a hostPath for pseudo-persistence.

```suggestion
          hostPath:
            path: /snekbox_dep_cache
            type: Directory
```

jb3 (changes requested):

jb3 (approved):
Looks good!

Issue #73, opened 2020-08-20
Rewrite dev/CI shell scripts in Python


MarkKoz:
Shell scripts may be fun in a masochistic kind of way, but they feel fragile and are unpleasant to maintain. This is a Python project under the name of a Python community, so let's be sane and use Python for these scripts.

PR #88, opened 2021-01-11
Use protobuf to parse the nsjail config and remove redundancies

### LINKED ISSUE #74: Redundancy regarding NsJail config values
I dislike that the following code exists when it is already defined in the config

https://github.com/python-discord/snekbox/blob/7a212edd8357d24e2dd88bc455418aaa895bc3c8/snekbox/nsjail.py#L23-L24
https://github.com/python-discord/snekbox/blob/7a212edd8357d24e2dd88bc455418aaa895bc3c8/snekbox/nsjail.py#L28
https://github.com/python-discord/snekbox/blob/7a212edd8357d24e2dd88bc455418aaa895bc3c8/scripts/.profile#L2
https://github.com/python-discord/snekbox/blob/7a212edd8357d24e2dd88bc455418aaa895bc3c8/scripts/.profile#L12-L15

These values are at risk of falling out of sync with the config. This applies more to `MEM_MAX`, since cgroup paths would be odd to modify.

Some ideas for solutions:

* Write a script to read the config and parse it into something Python can interact with.
    * The config format can ostensibly be parsed with simple regex, but I don't know how much I can trust its accuracy.
* Parse the file with protobuf, just like NsJail
    * This requires an extra dependency.
    * Their [config.proto](https://github.com/google/nsjail/blob/master/config.proto) file is needed somehow.
        * Copying it over means that it has to be kept in sync with whatever NsJail version is being used in the Docker image.
        * Using a submodule seems overkill and kind of annoying to use, but would solve the problem with simply copying over the file. The submodule could be used to build the Docker image rather than cloning the repo again when building. That would ensure the submodule version is kept in sync with the Docker image.

Either way, some sort of command line utility would be needed so the `.profile` shell script can access these values. To avoid that, `.profile` would need to just become a proxy to another Python script, which means effectively rewriting it using Python (see #73).

MarkKoz:
Fixes #74 

Protobuf is used to parse the config because it's the proper way to do it and it's less maintenance than a custom parser. A Python module is generated from the `protoc` tool. There's a simple shell script that gets the `config.proto` from NsJail's GitHub and invokes `protoc`. The Python module only needs to be re-generated when NsJail is updated.

I realise more shell scripts go against #73, but I plan to address that issue in a separate PR. It'll also address the NsJail version currently being hard-coded in separate places (the Python script and the Dockerfile).

I managed to remove one shell script anyway since it was practically necessary to reduce redundancy. It required using the code in snekbox to invoke NsJail. This has the benefit of all the code that parses the config and sets up cgroups. I could have imported it from the shell script, but I didn't want to write Python code in-line in the `.profile` script.

I'll also write unit tests later, since there may be more refactoring down the road. It's not worth it right now if it's just gonna have to get changed again.


jb3 (approved):
This all looks good, and works locally.

One thing I'd note for development is maybe the usage of the `script/protoc.sh` and a note about how the config parsing actually works/should be modified.

Other than that, thumbs up from me!

MarkKoz:
> One thing I'd note for development is maybe the usage of the script/protoc.sh

I plan to replace the script with on written in Python when I do #73. I'll document things better at that point.

>  and a note about how the config parsing actually works/should be modified.

Good point. I should document the config format, which is really just linking to the nsjail readme and maybe the protobuf file.

PR #90, opened 2021-01-12
 Fix caching of exposed packages

### LINKED ISSUE #89: Deployed pandas install is broken due to missing `six` dependency
https://discord.com/channels/267624335836053506/635950537262759947/798160975810920458

running `!e import pandas` results in an error message:

```
  File "/snekbox/user_base/lib/python3.9/site-packages/pandas/__init__.py", line 29, in <module>
    from pandas._libs import hashtable as _hashtable, lib as _lib, tslib as _tslib
  File "/snekbox/user_base/lib/python3.9/site-packages/pandas/_libs/__init__.py", line 13, in <module>
    from pandas._libs.interval import Interval
  File "pandas/_libs/interval.pyx", line 1, in init pandas._libs.interval
  File "pandas/_libs/hashtable.pyx", line 1, in init pandas._libs.hashtable
  File "pandas/_libs/missing.pyx", line 1, in init pandas._libs.missing
  File "/snekbox/user_base/lib/python3.9/site-packages/pandas/_libs/tslibs/__init__.py", line 30, in <module>
    from .conversion import OutOfBoundsTimedelta, localize_pydatetime
  File "pandas/_libs/tslibs/conversion.pyx", line 1, in init pandas._libs.tslibs.conversion
  File "pandas/_libs/tslibs/timezones.pyx", line 7, in init pandas._libs.tslibs.timezones
  File "/snekbox/user_base/lib/python3.9/site-packages/dateutil/tz/__init__.py", line 2, in <module>
    from .tz import *
  File "/snekbox/user_base/lib/python3.9/site-packages/dateutil/tz/tz.py", line 19, in <module>
    import six
ModuleNotFoundError: No module named 'six'

The above exception was the direct cause of the following exception:

  File "<string>", line 1, in <module>
  File "/snekbox/user_base/lib/python3.9/site-packages/pandas/__init__.py", line 33, in <module>
    raise ImportError(
ImportError: C extension: No module named 'six' not built. If you want to import pandas from the source directory, you may need to run 'python setup.py build_ext --force' to build the C extensions first.
```

MarkKoz:
Fix #89

The problem with `--ignore-installed` is that it also ignores packages in the target site, therefore re-installing all packages and making caching futile.

Install snekbox's dependencies to the default user site. During installation of the exposed packages, switch the user site to a custom one to ensure pip will not skip packages due to seeing them as already installed as one of snekbox's dependencies.

If pip made the `--root` option ignore packages outside the root, then using `--root` would be the best solution. There is a 5+ year old issue open about that.

There is also some refactoring of the Dockerfile.

### DIFF
--- a/.github/workflows/lint-test-build-push.yaml
+++ b/.github/workflows/lint-test-build-push.yaml
@@ -108,11 +108,12 @@ jobs:
           docker exec snekbox_test /bin/bash -c
           'apt-get -y update && apt-get install -y git=1:2.20.*'
 
+      # pre-commit's venv doesn't work with user installs.
       # Skip the flake8 hook because the following step will run it.
       - name: Run pre-commit hooks
         run: >-
           docker exec snekbox_test /bin/bash -c
-          'SKIP=flake8 pre-commit run --all-files'
+          'PIP_USER=0 SKIP=flake8 pre-commit run --all-files'
 
       # This runs `flake8` in the container and asks `flake8` to output
       # linting errors in the format of the command for registering workflow
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,6 +19,7 @@ RUN git clone \
 WORKDIR /nsjail
 RUN make
 
+# ------------------------------------------------------------------------------
 FROM python:3.9-slim-buster as base
 ENV PIP_NO_CACHE_DIR=false \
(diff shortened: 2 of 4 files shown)


Akarys42 (approved):
LGTM

PR #93, opened 2021-02-05
Fix NsJail Tests

MarkKoz:
#88 left tests in a broken state. This wasn't noticed since CI had a bug that prevented the step from failing despite tests failing. This PR fixes both of these issues.

### DIFF
--- a/.github/workflows/lint-test-build-push.yaml
+++ b/.github/workflows/lint-test-build-push.yaml
@@ -73,6 +73,7 @@ jobs:
           push: false
           load: true
           target: venv
+          build-args: DEV=1
           cache-from: |
             type=local,src=/tmp/.buildx-cache
             ghcr.io/python-discord/snekbox-base:latest
@@ -84,12 +85,6 @@ jobs:
         run: |
           export IMAGE_SUFFIX='-venv:${{ steps.sha_tag.outputs.tag }}'
           docker-compose up --no-build -d
-
-      # One of the unit tests needs to import numpy.
-      - name: Install dependencies
-        run: >-
-          docker exec snekbox_dev /bin/bash -c
-          'pipenv install --system --deploy --dev && pip install numpy'
 
       # Required by pre-commit.
       - name: Install git
--- a/.github/workflows/lint-test-build-push.yaml
+++ b/.github/workflows/lint-test-build-push.yaml
@@ -120,12 +120,15 @@ jobs:
         run: sudo swapoff -a
 
       # Run unittests and generate coverage report in the container
-      - name: Run unit tests and generate coverage report
+      - name: Run unit tests
         id: run_tests
         run: |
           echo '::set-output name=started::true'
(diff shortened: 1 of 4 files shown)


Akarys42 (approved):
LGTM! (except for the failing test)

MarkKoz:
Yeah this wasn't ready for review. I became too busy to continue and fix those issues.

.github/workflows/lint-test-build-push.yaml
```diff
@@ -150,8 +143,8 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          pip install coveralls~=2.1
-          coveralls
+          sudo pip install coveralls~=2.1
+          sudo coveralls
```
Akarys42, line 147: Is there any reason why we don't make a user install rather than a system install?

MarkKoz: IIRC it's cause the user site is not on PATH. It's more convenient to use `sudo` in this case.

MarkKoz: Though maybe it's bad practice, even in CI, and I should have kept using the Python version task...

MarkKoz: Don't yell at me, but I forced push to get rid of it. Turns out I had already dropped the commit locally last month, forgot about it, and commited changes over it.

MarkKoz:
@jb3 One of the tests is failing. This is a similar issue to what #94 was trying to fix, but the test still fails despite merging that into this branch.

```
======================================================================
ERROR: test_stdout_flood_results_in_graceful_sigterm (tests.test_nsjail.NsJailTests)
----------------------------------------------------------------------
  File "/home/runner/work/snekbox/snekbox/tests/test_nsjail.py", line 188, in test_stdout_flood_results_in_graceful_sigterm
    result = self.nsjail.python3(stdout_flood)
  File "/home/runner/work/snekbox/snekbox/snekbox/nsjail.py", line 236, in python3
    child.rmdir()
  File "/usr/local/lib/python3.9/pathlib.py", line 1352, in rmdir
    self._accessor.rmdir(self)
OSError: [Errno 16] Device or resource busy: '/sys/fs/cgroup/memory/snekbox-976324a6-c789-49eb-b88e-4702e4ac0626/NSJAIL.901'
```

Do you have any ideas?

.github/workflows/lint-test-build-push.yaml
Akarys42: Cool!

jb3:
So I guess the child cgroups are still busy. I'll check it out.

MarkKoz:
Turns out that killing the parent process does not kill its children. Therefore, we have nsjail's children (which are Python processes in this case) sticking around preventing child cgroups from being removed. This will be addressed in a separate PR.

PR #91, opened 2021-02-03
Dynamically create parent cgroups

jb3:
For some reason, when two memory intensive tasks were running in parallel, even while in child cgroups, the parent cgroup memory limit was tripping and sending an OOM kill to the child processes, even if theses processes were behaving reasonably under the 50MB limit. This issue is shown #83.

This PR resolves that by dynamically creating the parent cgroups which nsjail templates off, assigning them a random UUID name, prefixed by snekbox for clarity.

In testing I could not encounter the 137 error whilst running the snippets the issue author experienced, and still got the correct out of memory errors when operating over the 50 megabyte limit.

Closes #83.

PR #94, opened 2021-02-26
Remove all child cgroups before removing parents

jb3:
After #91 was merged we starting seeing errors with large stdout causing errors like the following:
```
2021-02-26 16:02:56,706 |    14 |                 snekbox.nsjail |     INFO | Executing code...
2021-02-26 16:02:56,741 |    14 |                 snekbox.nsjail |     INFO | Output exceeded the output limit, sending SIGKILL to NsJail.
2021-02-26 16:02:56,743 |    14 |                 snekbox.nsjail |     INFO | nsjail return code: 137
2021-02-26 16:02:56,744 |    14 |     snekbox.api.resources.eval |    ERROR | An exception occurred while trying to process the request
  File "/snekbox/snekbox/api/resources/eval.py", line 75, in on_post
    result = self.nsjail.python3(code)
  File "/snekbox/snekbox/nsjail.py", line 228, in python3
    Path(self.config.cgroup_mem_mount, cgroup).rmdir()
  File "/usr/local/lib/python3.9/pathlib.py", line 1352, in rmdir
    self._accessor.rmdir(self)
OSError: [Errno 16] Device or resource busy: '/sys/fs/cgroup/memory/snekbox-87fd27a1-d0e9-4387-8160-8d1a985d43e4'
```

I think that this external killing of nsjail does not let it clear up the child cgroup (which resides in a directory like `/sys/fs/cgroup/memory/snekbox-87fd27a1-d0e9-4387-8160-8d1a985d43e4/NSJAIL.12`). Once nsjail exits we then try to clear up the parent cgroup (in the example, `/sys/fs/cgroup/memory/snekbox-87fd27a1-d0e9-4387-8160-8d1a985d43e4`) which cannot succeed because the child cgroup exists.

This PR fixes that by removing the child cgroup before attempting to remove the parent, therefore recursively clearing the tree and ensuring that no cgroup can be busy.

My understanding is that it's safe to remove that cgroup (which I think technically removes the limits from nsjail) because by the point this code is reached the execution has either finished or has been killed.

### DIFF
--- a/snekbox/nsjail.py
+++ b/snekbox/nsjail.py
@@ -224,6 +224,18 @@ class NsJail:
 
         log.info(f"nsjail return code: {returncode}")
 
+        # If we hit a cgroup limit then there is a chance the nsjail cgroups did not
+        # get removed. If we don't remove them then when we try remove the parents
+        # we will get a "Device or resource busy" error.
+
+        children = []
+
+        children.extend(Path(self.config.cgroup_mem_mount, cgroup).glob("NSJAIL.*"))
+        children.extend(Path(self.config.cgroup_pids_mount, cgroup).glob("NSJAIL.*"))
+
+        for child in children:
+            child.rmdir()
+
         # Remove the dynamically created cgroups once we're done
         Path(self.config.cgroup_mem_mount, cgroup).rmdir()
         Path(self.config.cgroup_pids_mount, cgroup).rmdir()


jb3:
Also worth noting that the RAM limits and PID limits are not affected by this, only evaluations which result in a large stdout.

Screenshot 2021-02-26 at 17 07 18

Akarys42 (approved):
It does solve our current issue, thanks captain Joe!

Issue #83, opened 2020-12-07
Snekbox returning 137 when processing multiple memory-intensive requests simultaneously


nottheswimmer:
## Issue

I came across this issue when trying to run code that uses the sympy library invoked via a webserver in Snekbox. It seems that if there is a memory-intensive request, Snekbox will begin returning "137" out-of-memory errors if it receives multiple of them at the same time.

For an import to a large library like networkx, pandas, or sympy, merely two requests that import these at the same time will cause either one (typical) or both (rare) requests to fail.

## Environment

**Operating System**: Microsoft Windows 10 Education Version 10.0.18363 Build 18363
**Total Physical Memory**: 31.9 GB
**Docker version**: Docker version 19.03.13, build 4484c46d9d

## Steps to reproduce

1. Run snekbox

```sh
docker run --ipc=none --privileged -p 8060:8060 ghcr.io/python-discord/snekbox
```

2. Run the following code with one of the "CODE" lines uncommented. 

```python
import json
import urllib.request
from multiprocessing.dummy import Pool

PROCESSES = 3
# CODE = "import networkx"
# CODE = "import pandas"
# CODE = "import sympy"
CODE = "import time; ' ' * 33000000; time.sleep(0.1)"

def run_code_in_snekbox(code):
    body = {'input': code}
    req = urllib.request.Request("http://localhost:8060/eval")
    req.add_header('Content-Type', 'application/json; charset=utf-8')
    json_data = json.dumps(body).encode('utf-8')
    req.add_header('Content-Length', str(len(json_data)))
    with urllib.request.urlopen(req, json_data) as f:
        response = f.read()
    return json.loads(response.decode('utf-8'))


if __name__ == '__main__':
    arguments = [CODE] * PROCESSES
    with Pool(PROCESSES) as p:
        print(p.map(run_code_in_snekbox, arguments))
```

3. Observe that one request will succeed while the other two will fail:
```python
[{'stdout': '', 'returncode': 0}, {'stdout': '', 'returncode': 137}, {'stdout': '', 'returncode': 137}]
```

Container logs:
```
2020-12-07 00:57:27,616 |     7 |                 snekbox.nsjail |     INFO | Executing code...
2020-12-07 00:57:27,622 |     8 |                 snekbox.nsjail |     INFO | Executing code...
[...]

PR #95, opened 2021-03-08
Reparent orphaned child processors

jb3:
After receiving a set of busy resource errors after trying to remove dynamic cgroups introduced in #91, it became evident that Python interpreters were remaining active for a period after the nsjail parent had exited.

This PR fixes that by using [prctl](https://man7.org/linux/man-pages/man2/prctl.2.html) to register the Python NSJail workers as subreapers and collect orphaned processes.

Then, before cleaning up cgroups we use [`os.wait`](https://docs.python.org/3/library/os.html#os.wait) to ensure all processes have changed state before proceeding with the clean up.

A test for this PR is similar to that of the issue in #83, sending multiple requests for evaluation at the same time similar to:

```py
while True:
    print('abcdefghij')
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,6 +36,8 @@ RUN apt-get -y update \
         gcc=4:8.3.* \
         libnl-route-3-200=3.4.* \
         libprotobuf17=3.6.* \
+        libcap-dev=1:2.25-* \
+        libc6-dev=2.28-* \
     && rm -rf /var/lib/apt/lists/*
 RUN pip install pipenv==2020.11.15
 
--- a/Pipfile
+++ b/Pipfile
@@ -7,8 +7,9 @@ name = "pypi"
 falcon = "~= 2.0.0"
 gunicorn = "~= 20.0"
 jsonschema = "~= 3.2"
-protobuf = "~= 3.14"
+protobuf = "== 3.14.0"
 sentry-sdk = {extras = ["falcon"], version = "~= 0.19.3"}
+python-prctl = {version = "~=1.8",sys_platform = "== 'linux'"}
 
 [dev-packages]
 coverage = "~= 5.3"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "af33148cbd57da4f45c2ebf103ea0f466c6c7b518d316be30e8b2ef391102241"
+            "sha256": "34fc8f1bd3564988b6bdcd28178ea44bcb8a243b32417a81d31dc70dab523b8d"
         },
         "pipfile-spec": 6,
         "requires": {
@@ -89,7 +89,7 @@
                 "sha256:ecc33531a213eee22ad60e0e2aaea6c8ba0021f0cce35dbf0ab03dee6e2a23a1"
             ],
             "index": "pypi",
-            "version": "==3.14.0"
+            "version": "== 3.14.0"
         },
         "pyrsistent": {
             "hashes": [
@@ -97,6 +97,13 @@
             ],
             "markers": "python_version >= '3.5'",
             "version": "==0.17.3"
+        },
+        "python-prctl": {
+            "hashes": [
+                "sha256:b4ca9a25a7d4f1ace4fffd1f3a2e64ef5208fe05f929f3edd5e27081ca7e67ce"
+            ],
+            "markers": "sys_platform == 'linux'",
+            "version": "==1.8.1"
         },
         "sentry-sdk": {
             "extras": [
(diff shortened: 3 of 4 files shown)


MarkKoz:
`wait()` does indeed only wait for one child. However, what we can do is keep calling `wait()` in a loop until it raises a `ChildProcessError`, indicating there are no more unawaited child processes. This is safe according to the [documentation](https://man7.org/linux/man-pages/man2/wait.2.html#ERRORS); this error would not be raised for any other reason.

I am also concerned about the implications of being a subreaper. If we somehow became the parent of a process that shouldn't be killed immediately, then we may not be handling it properly. Granted, that doesn't seem possible since we don't launch any subprocessess besides nsjail currently. If it becomes a problem, we could look to something like [tini](https://github.com/krallin/tini) for inspiration.

MarkKoz:
The perplexing thing is that nsjail seems to handle all this for us in some (if not all) situations. This is what https://github.com/google/nsjail/commit/bb4e77686dfdae72dbaf1fa626519759f03de431 aimed to fix.

When it gets a fatal signal (not exactly sure on what this means), it does kill all child processes too. It seems like it keeps the loop going until there are no PIDs remaining, which accounts for children creating new processes while it is attempting to kill all children.

https://github.com/google/nsjail/blob/645eabd862e4eb20ec70a387fb7d50ecbc613f6e/nsjail.cc#L287-L291
https://github.com/google/nsjail/blob/645eabd862e4eb20ec70a387fb7d50ecbc613f6e/subproc.cc#L383-L392

This calls `reapProc` with `should_wait = true`.

https://github.com/google/nsjail/blob/645eabd862e4eb20ec70a387fb7d50ecbc613f6e/subproc.cc#L309-L317

The `wait4` call waits for the PID to terminate because `WNOHANG`  is _not_ set. If the return value equals the PID, it means the process changed state (i.e. terminated). Then it proceeds to clean up the cgroup via `finishFromParent`.

https://github.com/google/nsjail/blob/645eabd862e4eb20ec70a387fb7d50ecbc613f6e/cgroup.cc#L154-L173

Regardless of whether the signal is fatal or not, it first calls `reapProc(nsjconf_t* nsjconf)` and keeps doing that until there are no children remaining. This is a different `reapProc` which eventually calls the `reapProc(nsjconf_t* nsjconf, pid_t pid, bool should_wait)` described above.

https://github.com/google/nsjail/blob/645eabd862e4eb20ec70a387fb7d50ecbc613f6e/nsjail.cc#L275-L282
https://github.com/google/nsjail/blob/645eabd862e4eb20ec70a387fb7d50ecbc613f6e/subproc.cc#L342-L358

The `waitid` here doesn't actually wait for any processes. It's used to look for zombie processes (exited processes which haven't been `wait()`ed yet). It retrieves information about the signal used to terminate this process and stores it in `si` ([`siginfo_t`](https://www.mkssoftware.com/docs/man5/siginfo_t.5.asp)). `si_pid` is the PID of the process that sent the signal. It needs to use `si_pid` because `wait4` does not provide the PID of the process that received the signal (it doesn't know it directly either since it used `P_ALL` to check for _any_ child). This can still get the PIDs of the processes all the way at the bottom, since `si_pid` will be equal to the process's PID if the process kills itself.

I'm not sure why it breaks the loop if `si_pid` is 0. Maybe it's to avoid calling `wait4` with a PID of 0, since that is a special value which causes it to wait for any process in the same group as nsjail. I have seen some people talking about `si_pid` being 0, but there isn't a clear reason why that would happen.

Notice that `reapProc(nsjconf_t* nsjconf, pid_t pid, bool should_wait)` is not called with `should_wait = true`. Therefore, it is `false` and `wait4` does not suspend the calling thread to wait for the process state to change because `WNOHANG` is used. This means that if the process has already terminated, it will clean up the cgroup as described before (along with some logging) and it will also free up the resources of the process, which was a zombie until this point. If the process is still alive, then it will neither wait for it to terminate nor clean it up.

However, note that `reapProc(nsjconf_t* nsjconf)` is an infinite loop and so is `standaloneMode`. Both loops seem to stop only when no children remain. Therefore, I suppose it does effectively wait for children to terminate, but it does so via polling rather than suspending the thread. I suspect this is so that it can detect things like seccomp violations and exceeded time limits (though this is actually in a separate loop, so maybe not) as soon as possible. It will keep calling `reapProc(nsjconf_t* nsjconf, pid_t pid, bool should_wait)` for every child, and eventually that call will succeed for each child by way of `wait4` detecting the process has been terminated.

If my analysis is correct, I really don't understand why we're having to do any of this clean-up ourselves. We should be seeing this get logged if we set it to debug mode.

https://github.com/google/nsjail/blob/645eabd862e4eb20ec70a387fb7d50ecbc613f6e/cgroup.cc#L155

Can you take a look at the source and see if my understanding is correct or if you can find some reason for why cgroups or child processes wouldn't get cleaned up?

jb3:
Closing this since the solution (as discussed in #dev-core) appears to be sending SIGTERM to nsjail which will in turn distribute SIGKILL to any child processes.

PR #100, opened 2021-04-06
 Use PYTHONIOENCODING to enable utf-8 stdout for the nsjail pipe, and handle the potential case where this is bypassable

bast0006:
Since snekbox does not run with a tty, stdout is technically raw bytes, and thus incomplete surrogate pairs can be printed without the client application erroring, and instead fail within _consume_stdout when we attempt to decode it to a str.

This commit sets the PYTHONIOENCODING environment variable to inform python to open the pipe in utf-8 mode.

However, clever use of execl and os.unsetenv() can unset this environment variable, so we add a safety check to _consume_stdout to fail out of parsing output if it contains invalid unicode. This should only happen in deliberate cases, or significant bugs in python or a c library where output is printed to stdout ignoring the python stdout encoding.

### DIFF
--- a/snekbox/api/resources/eval.py
+++ b/snekbox/api/resources/eval.py
@@ -41,7 +41,7 @@ class EvalResource:
         The return codes mostly resemble those of a Unix shell. Some noteworthy cases:
 
         - None
-            The NsJail process failed to launch or return valid unicode output
+            The NsJail process failed to launch or the output was invalid Unicode
         - 137 (SIGKILL)
             Typically because NsJail killed the Python process due to time or memory constraints
         - 255
--- a/snekbox/api/resources/eval.py
+++ b/snekbox/api/resources/eval.py
@@ -41,7 +41,7 @@ class EvalResource:
         The return codes mostly resemble those of a Unix shell. Some noteworthy cases:
 
         - None
-            The NsJail process failed to launch
+            The NsJail process failed to launch or return valid unicode output
         - 137 (SIGKILL)
             Typically because NsJail killed the Python process due to time or memory constraints
         - 255
--- a/snekbox/nsjail.py
+++ b/snekbox/nsjail.py
@@ -208,7 +208,15 @@ class NsJail:
             except ValueError:
                 return CompletedProcess(args, None, "ValueError: embedded null byte", None)
 
-            output = self._consume_stdout(nsjail)
+            try:
+                output = self._consume_stdout(nsjail)
+            except UnicodeDecodeError:
+                return CompletedProcess(
+                    args,
+                    None,
+                    "UnicodeDecodeError: invalid unicode in output pipe",
+                    None,
+                )
 
             # When you send signal `N` to a subprocess to terminate it using Popen, it
             # will return `-N` as its exit code. As we normally get `N + 128` back, we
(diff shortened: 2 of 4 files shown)


snekbox/nsjail.py
```diff
@@ -208,7 +208,15 @@ def python3(self, code: str, *args) -> CompletedProcess:
...
                 return CompletedProcess(args, None, "ValueError: embedded null byte", None)
 
-            output = self._consume_stdout(nsjail)
+            try:
+                output = self._consume_stdout(nsjail)
+            except UnicodeDecodeError:
+                return CompletedProcess(
+                    args,
+                    None,
+                    "UnicodeDecodeError: invalid unicode in output pipe",
```
MarkKoz, line 217: Unicode should be capitalised.

tests/test_nsjail.py
```diff
@@ -100,6 +100,27 @@ def test_null_byte_value_error(self):
...
 
+    def test_print_bad_unicode_encode_error(self):
+        result = self.nsjail.python3("print(chr(56550))")
+        self.assertEqual(result.returncode, 1)
+        unicode_traceback = (
+            "Traceback (most recent call last):\n"
+            '  File "<string>", line 1, in <module>\n'
+            "UnicodeEncodeError: 'utf-8' codec can't encode character '\\udce6'"
+            " in position 0: surrogates not allowed\n"
+        )
```
MarkKoz, line 111: I'm not sure if asserting the exact traceback is the right approach to testing this. The way I see it, we don't care how Python deals with it as long as our program doesn't crash trying to get the output. In fact, do we even care if the exit code is 1?

MarkKoz (changes requested):
Please use `dedent` like the rest of the tests.

MarkKoz:
Also, this would need to be updated https://github.com/python-discord/snekbox/blob/main/snekbox/api/resources/eval.py#L43-L44

tests/test_nsjail.py
bast0006: I think we do care that it exits with 1, as that indicates the python process has successfully thrown an error message like we've expected. I was trying to find good ways of asserting that it simply returned a UnicodeEncodeError, as that's really what I wanted to check.

MarkKoz: On second thought, it's useful to assert the exit code to distinguish this case from the case of snekbox catching the error and handling it. Asserting the traceback is needed cause the exit code on its own would be too broad of an assertion. That being said, do you think simplifying it to `assert "UnicodeEncodeError" in output` would be sufficient?

bast0006: I think so. I'll go with that

snekbox/api/resources/eval.py
```diff
@@ -41,7 +41,7 @@ def on_post(self, req: falcon.Request, resp: falcon.Response) -> None:
         The return codes mostly resemble those of a Unix shell. Some noteworthy cases:
 
         - None
-            The NsJail process failed to launch
+            The NsJail process failed to launch or return valid unicode output
```
MarkKoz, line 44:
```suggestion
            The NsJail process failed to launch or the output was invalid Unicode
```

MarkKoz (approved):

MarkKoz:
Thanks @bast0006! Were you interested in working on adding support for byte output too?

bast0006:
I am, although I think we needed some bikeshedding on what exactly we were going to do there. I'll open an issue?

MarkKoz:
@bast0006 I opened one.

PR #108, opened 2021-05-17
feat: Allow custom arguments to be passed to eval

### LINKED ISSUE #105: Pass Custom Arguments to the Python Subprocess
Custom arguments to the Python subprocess would enable execution of other modules e.g. `timeit`. These arguments should be appended after the arguments in the nsjail config. In other words, they should replace the implicit `-c` argument

The arguments should be supplied as an array of strings in the JSON body of the POST request.

For example, given the current default configuration, passing `["-m", "timeit"]` should result in `/usr/local/bin/python -Squ -m timeit "the code here"`. Notice that `-c` is gone. Furthermore, `["-m", "timeit", "-n", "5", "-s", "x=5"]` should also work without problems.

ToxicKidz:
## Relevant Issues
Closes #105 

## Description
This PR makes it so that you can pass arguments, through an array of strings, that can replace the `"-c"` option. An example of how this would work would be:
```py
>>> import requests
>>> response = requests.post(                                                                                                   
...     "http://localhost:8060/eval",
...     json={"input": "[i for i in range(1000)]", "args": ["-m", "timeit"]}
... )
>>> response.json()
{'stdout': '10000 loops, best of 5: 24.5 usec per loop\n', 'returncode': 0}
```

### DIFF
--- a/snekbox/api/resources/eval.py
+++ b/snekbox/api/resources/eval.py
@@ -44,7 +44,9 @@ class EvalResource:
         """
         Evaluate Python code and return stdout, stderr, and the return code.
 
-        The optional `args` parameter can be passed, and it would replace the "-c" option.
+        A list of arguments for the Python subprocess can be specified as `args`.
+        Otherwise, the default argument "-c" is used to execute the input code.
+        The input code is always passed as the last argument to Python.
 
         The return codes mostly resemble those of a Unix shell. Some noteworthy cases:
 
--- a/snekbox/api/resources/eval.py
+++ b/snekbox/api/resources/eval.py
@@ -26,7 +26,7 @@ class EvalResource:
             },
             "args": {
                 "type": "array",
-                "contains": {
+                "items": {
                     "type": "string"
                 }
             }
@@ -44,6 +44,8 @@ class EvalResource:
         """
         Evaluate Python code and return stdout, stderr, and the return code.
 
+        The optional `args` parameter can be passed, and it woul replace the "-c" option.
+
         The return codes mostly resemble those of a Unix shell. Some noteworthy cases:
 
         - None
@@ -56,14 +58,14 @@ class EvalResource:
         Request body:
 
         >>> {
-        ...     "input": "print(1 + 1)"
+        ...     "input": "[i for i in range(1000)]",
         ...     "args": ["-m", "timeit"] # This is optional
         ... }
 
         Response format:
 
         >>> {
-        ...     "stdout": "2\\n",
+        ...     "10000 loops, best of 5: 23.8 usec per loop\n",
         ...     "returncode": 0
         ... }
 
@@ -80,7 +82,7 @@ class EvalResource:
         args = req.media.get("args", ("-c",))
 
         try:
-            result = self.nsjail.python3(code, extra_args=args)
+            result = self.nsjail.python3(code, py_args=args)
         except Exception:
             log.exception("An exception occurred while trying to process the request")
             raise falcon.HTTPInternalServerError
--- a/snekbox/api/resources/eval.py
+++ b/snekbox/api/resources/eval.py
@@ -23,6 +23,12 @@ class EvalResource:
         "properties": {
             "input": {
                 "type": "string"
+            },
+            "args": {
+                "type": "array",
+                "contains": {
+                    "type": "string"
+                }
             }
         },
         "required": [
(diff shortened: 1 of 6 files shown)


snekbox/api/resources/eval.py
```diff
@@ -51,6 +57,7 @@ def on_post(self, req: falcon.Request, resp: falcon.Response) -> None:
 
         >>> {
         ...     "input": "print(1 + 1)"
+        ...     "args": ["-m", "timeit"] # This is optional
```
MarkKoz, line 60: There needs to be a comma between the kv pairs.

MarkKoz: The response no longer matches the request.

snekbox/nsjail.py
```diff
@@ -171,12 +171,15 @@ def _consume_stdout(nsjail: subprocess.Popen) -> str:
 
         return "".join(output)
 
-    def python3(self, code: str, *args) -> CompletedProcess:
+    def python3(self, code: str, *args, extra_args: Iterable[str] = ("-c",)) -> CompletedProcess:
```
MarkKoz, line 174: The naming is currently too ambiguous. I think `*args` should be `nsjail_args` (no longer varargs) and `extra_args` should be `py_args`.

MarkKoz: Please add a test for the Python arguments. Now that I think about it, the nsjail args also lack a test, so feel free to add one for them too.

In this case, it may be better to mock the `subprocess.run` call and check the arguments it's called with rather than checking the result of the evaluation, unless you have a better idea.

snekbox/api/resources/eval.py
MarkKoz: Add a brief description of the `args` parameter to the docstring. I'm concerned that merely an example would not be sufficiently clear.

MarkKoz (changes requested):

snekbox/nsjail.py
```diff
@@ -171,12 +171,21 @@ def _consume_stdout(nsjail: subprocess.Popen) -> str:
...
+        self,
+        code: str,
+        *,
+        nsjail_args: Iterable[str] = (),
+        py_args: Iterable[str] = ("-c",)
+    ) -> CompletedProcess:
         """
         Execute Python 3 code in an isolated environment and return the completed process.
 
         Additional arguments passed will be used to override the values in the NsJail config.
```
MarkKoz, line 184: Revise this to explicitly mention `nsjail_args` rather than "additional arguments".

snekbox/nsjail.py
```diff
@@ -171,12 +171,21 @@ def _consume_stdout(nsjail: subprocess.Popen) -> str:
...
         These arguments are only options for NsJail; they do not affect Python's arguments.
+
+        The `py_args` keyword argument can be given, and this would replace the "-c" argument
+        given by default.
```
MarkKoz, line 188:
```suggestion
        `py_args` are arguments to pass to the Python subprocess before the code, 
        which is the last argument. By default, it's "-c", which executes the code given.
```

snekbox/api/resources/eval.py
```diff
@@ -38,6 +44,8 @@ def on_post(self, req: falcon.Request, resp: falcon.Response) -> None:
         """
         Evaluate Python code and return stdout, stderr, and the return code.
 
+        The optional `args` parameter can be passed, and it would replace the "-c" option.
```
MarkKoz, line 47: This should be clearer

```suggestion
        A list of arguments for the Python subprocess can be specified as `args`.
        Otherwise, the default argument "-c" is used to execute the input code.
        The input code is always passed as the last argument to Python.
```

MarkKoz (changes requested):

tests/api/test_eval.py
```diff
@@ -25,6 +25,24 @@ def test_post_invalid_schema_400(self):
...
+        self.assertEqual(input_result.status_code, 400)
+        self.assertEqual(args_result.status_code, 400)
+
+        expected = {
+            "title": "Request data failed validation",
+            "description": "400 is not of type 'string'"
+        }
+
+        self.assertEqual(expected, input_result.json)
+        self.assertEqual(expected, args_result.json)
```
MarkKoz, line 44: Use a subtest context manager to isolate the assertions. Furthermore, it could use a for loop over the JSON bodies to reduce code redundancy.

MarkKoz (changes requested):

MarkKoz (approved):

PR #112, opened 2021-07-17
Increase available PIDs to 5

jb3:
This PR increases the PIDs cgroup limit to 5. Processes spawned in snekbox will have up to 5 PIDs available, each sharing the same memory limits and environment as the parent python process. As far as I could see in testing this does appear safe and processes behave as expected, even when detatching from the parent and so on.

I'd appreciate if anyone with a local setup could give this a shot and play with things like multiprocessing, subprocess and any other low level proc interfaces.

### DIFF
--- a/config/snekbox.cfg
+++ b/config/snekbox.cfg
@@ -9,11 +9,11 @@ time_limit: 6
 
 keep_env: false
 envar: "LANG=en_US.UTF-8"
-envar: "OMP_NUM_THREADS=1"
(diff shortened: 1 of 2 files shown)


jchristgit (approved):
As long as this goes into the same cgroup, this is good, Joe, very good

MarkKoz:
Thank @jb3. Can you think of any way to write a test for the resource limits being shared?

MarkKoz (approved):
The tests are passing, so I'm gonna go off that and say this works.

PR #113, opened 2021-07-21
Enable a limited tmpfs for shared memory

jb3:
This PR introduces a small 40mb tmpfs mounted at /dev/shm to allow shared memory, utilised by `multiprrocessing` to work. This means that `multiprocessing.Pool` should work in snekbox (and we have a new test that verifies this).

I'd appreciate if anyone would like to test that you try break the 40mb limit and other general trickery surrounding these new capabilities.

### DIFF
--- a/config/snekbox.cfg
+++ b/config/snekbox.cfg
@@ -106,6 +106,14 @@ mount {
     rw: false
 }
 
+mount {
+    dst: "/dev/shm"
+    fstype: "tmpfs"
+    rw: true
+    is_bind: false
+    options: "size=40m"
+}
+
 cgroup_mem_max: 52428800
 cgroup_mem_swap_max: 0
 cgroup_mem_mount: "/sys/fs/cgroup/memory"
--- a/tests/test_nsjail.py
+++ b/tests/test_nsjail.py
@@ -23,6 +23,9 @@ class NsJailTests(unittest.TestCase):
         self.logger = logging.getLogger("snekbox.nsjail")
         self.logger.setLevel(logging.WARNING)
 
+        # Hard-coded because it's non-trivial to parse the mount options.
+        self.shm_mount_size = 40 * Size.MiB
+
     def eval_code(self, code: str):
         return self.nsjail.python3(["-c", code])
 
@@ -124,6 +127,25 @@ class NsJailTests(unittest.TestCase):
         exit_codes = result.stdout.strip().split()
         self.assertIn("-9", exit_codes)
         self.assertEqual(result.stderr, None)
+
+    def test_multiprocessing_pool(self):
+        # Validates that shm is working as expected
+        code = dedent(
+            """
+            from multiprocessing import Pool
+
+            def f(x):
+                return x*x
+
+            with Pool(2) as p:
+                print(p.map(f, [1, 2, 3]))
+        """
+        )
+
+        result = self.eval_file(code)
+
+        self.assertEqual(result.stdout, "[1, 4, 9]\n")
+        self.assertEqual(result.returncode, 0)
 
     def test_read_only_file_system(self):
         for path in ("/", "/etc", "/lib", "/lib64", "/snekbox", "/usr"):
@@ -335,35 +357,60 @@ class NsJailTests(unittest.TestCase):
             log.output,
         )
 
-    def test_shm_and_tmp_not_mounted(self):
-        for path in ("/dev/shm", "/run/shm", "/tmp"):
-            with self.subTest(path=path):
-                code = dedent(
-                    f"""
-                    with open('{path}/test', 'wb') as file:
-                        file.write(bytes([255]))
-                    """
-                ).strip()
-
-                result = self.eval_file(code)
-                self.assertEqual(result.returncode, 1)
-                self.assertIn("No such file or directory", result.stdout)
-                self.assertEqual(result.stderr, None)
-
-    def test_multiprocessing_shared_memory_disabled(self):
+    def test_tmp_not_mounted(self):
         code = dedent(
             """
-            from multiprocessing.shared_memory import SharedMemory
-            try:
-                SharedMemory('test', create=True, size=16)
-            except FileExistsError:
-                pass
-            """
+            with open('/tmp/test', 'wb') as file:
+                file.write(bytes([255]))
+        """
         ).strip()
 
         result = self.eval_file(code)
         self.assertEqual(result.returncode, 1)
-        self.assertIn("Function not implemented", result.stdout)
+        self.assertIn("No such file or directory", result.stdout)
+        self.assertEqual(result.stderr, None)
+
+    def test_multiprocessing_shared_memory(self):
+        cases = (
+            (self.shm_mount_size, self.shm_mount_size, 0),
+            # Even if the shared memory object is larger than the mount,
+            # writing data within the size of the mount should succeed.
+            (self.shm_mount_size + 1, self.shm_mount_size, 0),
+            (self.shm_mount_size + 1, self.shm_mount_size + 1, 135),
+        )
+
+        for shm_size, buffer_size, return_code in cases:
+            with self.subTest(shm_size=shm_size, buffer_size=buffer_size):
+                # Need enough memory for buffer and bytearray plus some overhead.
+                mem_max = (buffer_size * 2) + (400 * Size.MiB)
+                code = dedent(
+                    f"""
+                    from multiprocessing.shared_memory import SharedMemory
+
+                    shm = SharedMemory(create=True, size={shm_size})
+                    shm.buf[:{buffer_size}] = bytearray([1] * {buffer_size})
+                """
+                ).strip()
+
+                result = self.eval_file(code, nsjail_args=("--cgroup_mem_max", str(mem_max)))
+
+                self.assertEqual(result.returncode, return_code)
+                self.assertEqual(result.stdout, "")
+                self.assertEqual(result.stderr, None)
+
+    def test_multiprocessing_shared_memory_mmap_limited(self):
+        """The mmap call should be OOM trying to map a large & sparse shared memory object."""
+        code = dedent(
+            f"""
+            from multiprocessing.shared_memory import SharedMemory
+
+            SharedMemory(create=True, size={self.nsjail.config.cgroup_mem_max + Size.GiB})
+        """
+        ).strip()
+
+        result = self.eval_file(code)
+        self.assertEqual(result.returncode, 1)
+        self.assertIn("[Errno 12] Cannot allocate memory", result.stdout)
         self.assertEqual(result.stderr, None)
 
     def test_numpy_import(self):


tests/test_nsjail.py
```diff
@@ -95,6 +95,24 @@ def f():
...
 
+    def test_multiprocessing_pool(self):
+        # Validates that shm is working as expected
+        code = dedent("""
+            from multiprocessing import Pool
+
+            def f(x):
+                return x*x
+
+            if __name__ == '__main__':
```
MarkKoz, line 106: This check isn't necessary in this context.

jchristgit (approved):
This looks good to me.

As for your request to test breaking the memory limit, writing there using `multiprocessing.shared_memory` would have also been my one and only test, and since we're doing that here, I personally can't think of another possibility to break this. Since it says "File too large", what happens if multiple `shared_memory.SharedMemory` objects are created that exceed the limit together? I would assume that the first file exceeding the limit hit the error?

config/snekbox.cfg
```diff
@@ -104,10 +104,18 @@ mount {
...
+    rw: true
+    is_bind: false
+    options: "size=40m"
+}
+
 cgroup_mem_max: 52428800
 cgroup_mem_mount: "/sys/fs/cgroup/memory"
 
-cgroup_pids_max: 5
+cgroup_pids_max: 6
```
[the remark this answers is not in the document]
MarkKoz: We need to understand exactly why it has to be upped to 6. Someone showed that it spawns some extra internal threads as an implementation detail. The rest of the threads are based on `os.cpu_count` if the count is unspecified.

How consistent and reliable is that behaviour? Furthermore, should we be using `os.cpu_count` to automatically calculate what this value should be? Such feature would have to be optional, otherwise it'd force an effective minimum on the PID limit, which not all users may want.

MarkKoz (changes requested):

config/snekbox.cfg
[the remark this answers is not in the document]
MarkKoz: On second thought, using `cpu_count` for calculation sounds like tight coupling, especially considering it's also relying on an implementation detail of the number of threads the pool creates. I suppose hard-coding the PID limit in the config is fine, as long as we can ensure the `cpu_count` is consistent in the production environment.

MarkKoz:
Someone made me realise that shared memory could be accessed across NsJail instances. Can you think of a way have isolated mounts for each instance?

jb3:
How could that happen? We create a new tmpfs for every nsjail instance — it's not shared.

MarkKoz:
> How could that happen? We create a new tmpfs for every nsjail instance — it's not shared.

Is it really a new one or is it just mounting the same directory over and over?

MarkKoz:
Is that what is_bind false does?

jb3:
> Is it really a new one or is it just mounting the same directory over and over?

https://github.com/python-discord/snekbox/blob/e6cf1c4dce96e45313f125473dc30582f4d443ef/config/snekbox.cfg#L107-L113

If this works the same way that it would in the equivalent fstab, it will be a new tmpfs instance for each execution. I'm having trouble reading through how nsjail actually allocates this though, I can't tell if it is doing something special or just mounting a filesystem in the typical way.

> Is that what is_bind false does?

is_bind enables:
- `MS_BIND` ("A bind mount makes a file or a directory subtree visible at another point within the single directory hierarchy")
- `MS_REC` ("Used in conjunction with MS_BIND to create a recursive bind mount, and in conjunction with the propagation type flags to recursively change the propagation type of all of the mounts in a subtree.")
- `MS_PRIVATE` ("Make this mount point private.  Mount and unmount events do not propagate into or out of this mount point.")

I'm not sure any of these are required, it might be a dud option for tmpfs mounts.

MarkKoz:
Are these standard semantics for tmpfs? For the sake of my curiosity, is it even possible to mount the same tmpfs under different directories?

jb3:
You can achieve that through symlinking, I don't think there is a "same tmpfs" due to the nature of tmpfs allocation (a new tmpfs for every mountpoint, as far as I understand).

MarkKoz:
Okay, great. Just need to address the comments I left a couple weeks ago.

tests/test_nsjail.py
MarkKoz: a945f06

tests/test_nsjail.py
```diff
@@ -184,31 +202,27 @@ def test_log_parser(self):
...
+    def test_multiprocessing_shared_memory_limited(self):
         code = dedent("""
-            from multiprocessing.shared_memory import SharedMemory
-            try:
-                SharedMemory('test', create=True, size=16)
-            except FileExistsError:
-                pass
+            from multiprocessing import shared_memory
+            shm_a = shared_memory.SharedMemory(create=True, size=45_000_000)
+            shm_a.buf[:45_000_000] = bytearray([1] * 45_000_000)
```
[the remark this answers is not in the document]
MarkKoz: On second thought, screw that. Complexity of parsing outweighs the minor benefit.

tests/test_nsjail.py
```diff
@@ -184,31 +200,27 @@ def test_log_parser(self):
...
         """).strip()
 
         result = self.nsjail.python3(code)
         self.assertEqual(result.returncode, 1)
-        self.assertIn("Function not implemented", result.stdout)
+        self.assertIn("File too large", result.stdout)
```
wookie184, line 223: The file size limit is a different thing to the tmpfs size, and I think this is just testing the file size limit (`rlimit_fsize` parameter in nsjail, set to 1mb by default). To test the tmpfs size we'd have to create a bunch of files below 1mb. We should then get a different error, `OSError: [Errno 28] No space left on device`

tests/test_nsjail.py
```diff
@@ -48,11 +48,10 @@ def test_memory_returns_137(self):
         self.assertEqual(result.stderr, None)
 
     def test_subprocess_resource_unavailable(self):
-        code = dedent("""
+        code = dedent(f"""
             import subprocess
 
-            # Max PIDs is 5.
-            for _ in range(6):
+            for _ in range({self.nsjail.config.cgroup_pids_max + 1}):
```
wookie184, line 54: I guess we could not include the `+ 1`, since the current process will count as 1, although it doesn't really matter.

wookie184 (changes requested):
Thoughts on adding a test that creates lots of empty directories to ensure there are limits on that? Memory taken up by them doesn't count towards the tmpfs size afaict. Currently the limit seems to be due to it hitting `cgroup_mem_max`, which is fine, although we could also consider setting `nr_inodes` in the mount options to explicitly limit that if we wanted to.

tests/test_nsjail.py
MarkKoz: I've picked up this PR again, and in my testing, the process crashes with code 135 (`SIGBUS`). To trigger this, the code must write to the shared memory buffer. Merely creating the shared memory object is not enough.

What I've been stuck on is understanding what's going on when just the shared memory object is created. [Here](https://github.com/python/cpython/blob/main/Lib/multiprocessing/shared_memory.py#L104-L115) is the CPython code for it.

First it calls `shm_open`, and it has the `O_CREAT` flag set since `create=True` is passed in this test. This initially creates an empty file as documented [here](https://www.man7.org/linux/man-pages/man3/shm_open.3.html):

> A new shared memory object initially has zero length—the size of the object can be set using [ftruncate(2)](https://www.man7.org/linux/man-pages/man2/ftruncate.2.html).  The newly allocated bytes of a shared memory object are automatically initialized to 0.

Indeed, Python afterwards calls `ftruncate` with the size. This will set the file to exactly the given size (in this case, it has to extend the file from a size of 0 to the given size).

> The truncate() and ftruncate() functions cause the regular file named by path or referenced by fd to be truncated to a size of precisely length bytes.

Indeed, when I `stat` the file, I can see `st_size` is set to the expected size. Finally, it `mmap`s the entire file.

The strange thing is that despite the cgroup memory limit being in effect, it's possible to create files in `/dev/shm` much larger than the memory limit. For example, with a memory limit of `52428800`, I can create a file 13 times that size:

```
python -m snekbox 'from multiprocessing.shared_memory import SharedMemory; shm = SharedMemory(create=True, size=52428800 * 13); import os; print(os.stat("/dev/shm/" + shm.name))' --time_limit 0

os.stat_result(st_mode=33152, st_ino=2, st_dev=60, st_nlink=1, st_uid=65534, st_gid=65534, st_size=681574400, st_atime=1693161926, st_mtime=1693161926, st_ctime=1693161926)
```

However, when I inspect the memory in the cgroup (while the jailed process is running with a sleep), it is much smaller than the size of the file.

```
cat /sys/fs/cgroup/NSJAIL.92/memory.current 
14569472
```
```
cat /sys/fs/cgroup/NSJAIL.92/memory.stat | grep shm
shmem 0
shmem_thp 0
```
```
cat /sys/fs/cgroup/NSJAIL.100/memory.stat | grep map
file_mapped 0
```

The [docs](https://git.kernel.org/pub/scm/linux/kernel/git/tj/cgroup.git/tree/Documentation/admin-guide/cgroup-v2.rst#n1397) state that `shmem` in `memory.stat` should include shared memory and anything backed by tmpfs:

> shmem
> Amount of cached filesystem data that is swap-backed, such as tmpfs, shm segments, shared anonymous mmap()s

It seems that the cgroup memory controller is unable to capture memory usage from shared memory, which is a concern.

If the shared memory size is too large (unclear how large is too large), at least the `mmap` will fail:

```
  File "<string>", line 1, in <module>
  File "/usr/local/lib/python3.11/multiprocessing/shared_memory.py", line 115, in __init__
    self._mmap = mmap.mmap(self._fd, size)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^
OSError: [Errno 12] Cannot allocate memory
```

MarkKoz: It seems like no memory is consumed by just using `ftruncate`:

```
python -m snekbox '
import os
import _posixshmem
try:
   fd = _posixshmem.shm_open("test", os.O_CREAT | os.O_EXCL | os.O_RDWR, mode=0o600)
   os.ftruncate(fd, 52428800 * 5000000)
   print(os.stat(fd))
finally:
   _posixshmem.shm_unlink("test")
' --time_limit 0
```

```
os.stat_result(st_mode=33152, st_ino=2, st_dev=60, st_nlink=1, st_uid=65534, st_gid=65534, st_size=262144000000000, st_atime=1693165871, st_mtime=1693165871, st_ctime=1693165871)
```

That's 262 TB which would definitely run out of memory (or swap/disk space if that was used).

MarkKoz: Linux has a concept of "sparse files", so `fallocate` does not actually cause memory to be allocated. Thus, there is no concern about being OOM here with insanely large `fallocate` calls. See https://stackoverflow.com/questions/58571685/do-files-in-dev-shm-take-up-memory-when-grown-with-ftruncate-but-are-not-writte

tests/test_nsjail.py
MarkKoz: Indeed. I've removed this change in the recent rebase.

tests/test_nsjail.py
MarkKoz: I've updated the tests to address the initial comment https://github.com/python-discord/snekbox/pull/113/commits/8028c08514a507e3de82a1260cbc79fe34095ccd, and I also bottomed out on the other thing I was confused on.

config/snekbox.cfg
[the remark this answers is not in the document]
MarkKoz: The test was changed to read this value from the config here c50b5473d74d6d2ef6f29eca68c65b4f61b3466d

tests/test_nsjail.py
```diff
@@ -335,35 +357,60 @@ def test_log_parser(self):
...
+            # Even if the shared memory object is larger than the mount,
+            # writing data within the size of the mount should succeed.
+            (self.shm_mount_size + 1, self.shm_mount_size, 0),
+            (self.shm_mount_size + 1, self.shm_mount_size + 1, 135),
+        )
+
+        for shm_size, buffer_size, return_code in cases:
+            with self.subTest(shm_size=shm_size, buffer_size=buffer_size):
+                # Need enough memory for buffer and bytearray plus some overhead.
+                mem_max = (buffer_size * 2) + (400 * Size.MiB)
```
MarkKoz, line 385: I don't know why it needs so much overhead 🤷 but in any case, the point of this test isn't to test the normal NsJail OOM killer — there are other tests for that.

MarkKoz:
@wookie184 Testing for inode limits (and having inode limits) sounds like a good idea. Do you mind writing up a new issue for that? I think that situation would also apply to our temporary file system feature.

wookie184 (approved):
Great stuff, thanks for working on this

PR #117, opened 2021-10-11
Bump To Python 3.10

HassanAbouelela:
This PR bumps the python version of the project to 3.10.

From my testing, everything builds as expected. All functions work (the long life of the other 3.10 snekbox bot helps support that), and what isn't available can be built since we have gcc in our docker configuration.

I also went through the deployment packages and cleaned them up a bit. Everything (except scipy) worked straight away, but I still did the following:

- For most dependencies, the current version has explicit or implicit support for 3.10
- For some dependencies, I had to bump version to get 3.10 support

A few dependencies were not upgraded, and some removed:

Not updated - no wheels or explicit support for 3.10, but they still work:
- pendulum
- PyYAML

removed:
- python-levenshtein - Ther version we were on was yanked, and it seems unmaintained. Since it's not used that often, I decided to remove it
- scipy - Explicitly lacks support for 3.10 at the moment.

### DIFF
--- a/.github/workflows/lint-test-build-push.yaml
+++ b/.github/workflows/lint-test-build-push.yaml
@@ -140,7 +140,7 @@ jobs:
         id: python
         uses: actions/setup-python@v2
         with:
-          python-version: '3.9'
+          python-version: '3.10'
 
       # We'll only ever need a single dependency in this python
       # environment and we'll only use it in the CI, so let's
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -2,7 +2,7 @@
 
 ## Initial Setup
 
-A Python 3.9 interpreter and the [pipenv] package are required. Once those requirements are satisfied, install the project's dependencies:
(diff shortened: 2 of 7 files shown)


HassanAbouelela:
Having a weird thing with coverage, where it seems to think a branch was not hit when it was supposed to. That is theoretically possible, but the code and tests did not change, so I don't think there is an actual decrease in coverage, just either a bug with coverage now, or the one used to generate the initial report. I do not think this is related to my changes.

HassanAbouelela:
I'll leave this here as a historical footnote:

The latest 5.x version of coverage.py seems to have changed/broken something that meant we ended up with a lower coverage % than we did on main when it was last published. Version 6.x fixes that problem, but does not appear to be compatible with coveralls.

With no clear fix available, Mark suggested we remove the failure on coverage decreases, which is exactly what was done through the coveralls site. The coverage will technically be incorrect, but now it shouldn't block anything anymore.

PR #119, opened 2021-11-12
Fix Tests On Protobuf 3.19+

HassanAbouelela:
Temporary PR to test the actions.


MarkKoz:
The nsjail version being used is 2.9, not 3.0, unless you plan on upgrading that too.

HassanAbouelela:
Well, this is interesting. It seems changing the generated config file from its default name now breaks sneakbox. I'm struggling to investigate this much, especially because I can't reproduce locally. I've tested it here in actions (if you couldn't already tell from the 10 force pushes), and it seems that no combination of nsjail config name and final config name works besides the default config_pb2.

I haven't had a chance to investigate yet, but it seems everything is working with the rename. I'll try looking into it a bit more, but the errors are currently very mysterious, and rename might just be the least headache.

MarkKoz:
I'm not sure how it ever worked to begin with, but even as far back as 2016 there was a report of this same error when the prefix was removed (so it's not something introduced by a new protoc version). See https://github.com/tgalal/python-axolotl/issues/14 Even the generated file in the main branch has the original module name hard-coded at the end of the file.

There isn't a way to tell protoc to use a different file name: https://github.com/protocolbuffers/protobuf/issues/2283; https://github.com/protocolbuffers/protobuf/issues/7061

Maybe if you change the `__module__` constants it will work with the new name?

HassanAbouelela:
Changing the name of the `config.proto` file used to generate our config changes the `__module__` constant. I've tested various combinations of `__module__` and file names, and it seemed nothing besides `config_pb2` works for this.

These are all the combinations I've tried besides the current one, and they all failed with the same exception:

`Snekbox config` & `nsjail config`:
- [`proto_config` and `proto_config`](https://github.com/python-discord/snekbox/compare/5e618f3...cbfdf4842f63)
- [`proto_config` and `config`](https://github.com/python-discord/snekbox/compare/5e618f3...8d54a50bcb95)
- [`config` and `x` (just something random)](https://github.com/python-discord/snekbox/compare/5e618f3...64e2a1ac2460)

I'm not sure if `config_pb2` is just hardcoded somewhere.

MarkKoz:
By `__module__` I mean the key in the dictionary here (there are several instances) https://github.com/python-discord/snekbox/blob/fbc9476cf911ee3e3fc1c2d30cf3702a46f2b3ef/snekbox/config_pb2.py#L995

I am suggesting that changing this value to `config` after the file is generated may fix it. But if it doesn't work then I can live with `config_pb2`.

HassanAbouelela:
I think that's what we had before I started messing around with the file, but I tried it again in c00c4cf. Still seeing the same error.

MarkKoz:
Ah okay I guess it's just doomed to fail if it's renamed

MarkKoz:
I just started getting the same error about type descriptors in my local branch which is based off master. I'm not sure what triggered it. It seems to only happen when I run it through docker-compose up`. If I use `pipenv run devsh` and then `python -m snekbox`, it works.

Will you be able to finish the PR soon? Let's just agree to not bother renaming the file since it breaks things.

---

Also, please target the current ref in the master branch rather than 3.0. Since 3.0, a feature has been added to tell NSJAIL to set the swap memory limit, so snekbox won't have to do that any more.

jb3 (approved):

PR #121, opened 2021-12-07
Add the --use_cgroupv2 flag when relevant

Akarys42:
According to https://github.com/google/nsjail/pull/119, the flag should be passed for NsJail to try to use cgroupv2. This commit will use the `/sys/fs/cgroup` structure to guess the installed version, and depending on the version add that flag.

I couldn't confirm this actually solves the issue with v2 but at least it doesn't break with v1.

### DIFF
--- a/snekbox/nsjail.py
+++ b/snekbox/nsjail.py
@@ -44,10 +44,12 @@ class NsJail:
     def __init__(self, nsjail_binary: str = NSJAIL_PATH):
         self.nsjail_binary = nsjail_binary
         self.config = self._read_config()
-    
+
+        log.info(f"Cgroups version: {self._probe_cgroup_version()}")
+
     @staticmethod
     def _probe_cgroup_version() -> int:
-        """Poll the filesystem and return the guessed cgroup version"""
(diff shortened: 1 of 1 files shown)


HassanAbouelela (approved):
Confirmed working on cgroups 1

snekbox/nsjail.py
```diff
@@ -42,6 +45,19 @@ def __init__(self, nsjail_binary: str = NSJAIL_PATH):
         self.nsjail_binary = nsjail_binary
         self.config = self._read_config()
 
+        log.info(f"Cgroups version: {self._probe_cgroup_version()}")
+
+    @staticmethod
+    def _probe_cgroup_version() -> int:
+        """Poll the filesystem and return the guessed cgroup version."""
+        # Right now we check whenever the controller path exists
+        version = 2 if CGROUPV2_PROBE_PATH.exists() else 1
```
MarkKoz, line 54: Ideally this would use `statfs` to get the filesystem type, but Python doesn't make it easy to perform this syscall unfortunately.

MarkKoz (commented):
I suppose I am late but I'll review this anyway.

PR #127, opened 2021-12-22
Add cgroupv2 support

### LINKED ISSUE #83: Snekbox returning 137 when processing multiple memory-intensive requests simultaneously

### LINKED ISSUE #102: Support cgroup v2
NsJail supports v2 and ideally snekbox should too.

### LINKED ISSUE #125: Update NsJail to get support for setting the swap memory limit
https://github.com/google/nsjail/pull/186 was merged and it'd be helpful for snekbox. It would mean that at least for cgroupv2 support, snekbox wouldn't have to manually write 0 to `memory.swap.max`. This PR isn't available in any NsJail release, so the Dockerfile will have to clone based on commit hash.

### LINKED ISSUE #126: Cgroupv2 - cgroup.subtree_control is empty in the root cgroup and can't be populated
By default, Docker uses a private cgroup namespace when the host system uses cgroupv2. This results in the root cgroup within the container having an empty `cgroup.subtree_control`, which means the child cgroups NsJail creates will not have any controllers enabled. Attempting to write to the root `cgroup.subtree_control` results in a "device or resource busy" error, which seems to be because the cgroup already has processes in it (it's the root cgroup, after all). However, the exact cause for this error has not been confirmed.

`docker run` has a `--cgroupns` option which can be set to `host` to use the host's cgroup namespace instead of a private one. This works around the empty `cgroup.subtree_control` but at the cost of not having a private namespace. This is in fact the default behaviour when cgroupv1 is used. Another downside is that this option cannot be configured in the Docker Compose file (https://github.com/compose-spec/compose-spec/issues/148). To still use Docker Compose, this setting would need to be set globally via the `default-cgroupns-mode` Docker daemon option. Otherwise, the container would have to be started with `docker run` instead.

I looked into the `--cgroup-parent` option as well. I created a cgroup `/sys/fs/cgroup/NSJAIL.slice` and then enabled some controllers in its `cgroup.subtree_control` before starting the container. However, the cgroup Docker creates within `NSJAIL.slice` still ends up having an empty `cgroup.subtree_control`. That makes sense, since when I created `NSJAIL.slice`, `cgroup.subtree_control` also started out empty (as is documented by the various manpages on cgroupv2). I was hoping that `NSJAIL.slice` would become the root cgroup in the container, which is not the case.

As mentioned in the [cgroup namespaces manpage](https://man7.org/linux/man-pages/man7/cgroup_namespaces.7.html),

> When a process creates a new cgroup namespace using clone(2) or unshare(2) with the CLONE_NEWCGROUP flag, its current cgroups directories become the cgroup root directories of the new namespace.

However, Docker seems to create a new cgroup for each container with some hash in its name right before starting the container. Thus, there seems to be no opportunity to write to `cgroup.subtree_control` before the cgroup is populated with processes. Thus, the only solution I see currently is to rely on `--cgroupns host`.

MarkKoz:
NsJail has supported it for a while, and several other users have encountered lack of cgroupv2 support as an issue. However, Python Discord hasn't quite felt urgency get with the times until our Kubernetes cluster unexpectedly updated to Debian 11, which defaults to cgroupv2. Alas, our production instance of Snekbox was broken. Finally, the long-awaited support for cgroupv2 is here. 

Fixes #102 
Fixes #125 
Fixes #126 

There was a lot of rebasing involved. Sorry for the commits appearing out of order on GitHub (they are correctly ordered in `git log`).

# Changes

## NsJail

NsJail has been upgraded. This version doesn't have a proper release yet (i.e. it's not tagged). This update is beneficial because it adds support for telling NsJail to set the swap memory limit in the cgroup, meaning Snekbox no longer has to do that.

### v2

Cgroupv2 is supported. Some initialisation is done to ensure child cgroups created by NsJail will have controllers enabled. This doesn't necessarily reflect the controllers that are demanded by the configuration. Rather, it enables all possible controllers if none are enabled for children. Otherwise, it just uses whatever controllers were already enabled.

Both a host and private cgroup namespace are supported (as configurable by `docker run --cgroupns`). Private namespaces were a bit tricky; see #126 and `snekbox.utils.cgroup.init_v2` for details.

### v1

v1 has been simplified thanks to the NsJail update. It's no longer necessary to create separate parent cgroups for each NsJail process to fix #83. Thus, all Snekbox does now is create the memory and pids cgroup mounts if they don't exist.

Interestingly, I discovered that if e.g. `cgroup_mem_parent` is set to an empty string, then NsJail can just write directly to `cgroup_mem_mount`. Thus, it could write to the root of that controller and no parent even has to be created by Snekbox. However, I decided to leave that bit of code so that non-empty `cgroup_xxx_parent` are supported.

### Version Detection

The cgroup version is detected by examining the cgroup mount's directory structure. This relies on the NsJail mount settings being accurate; I didn't want to hard-code `sys/fs/cgroup`. It is technically possible to search for cgroup fs mounts by examining `/proc/mounts`, but I didn't want to introduce the additional complexity of parsing that. Furthermore, v1 is a bit trickier to detect since the root is a tmpfs. I didn't want to make assumptions, but I wasn't clear on was guarantees the spec makes regrading the structure of v1 and v2.

Version detection also takes into account the `use_cgroupv2` NsJail setting. If Snekbox can, it will override the configured setting with the version that it detects the system to actually have. It can do this for overriding v1 with v2, but not the other way around (because if `use_cgroupv2` already exists, there isn't a command-line argument that can negate it).  A warning is shown for the latter.

`use_cgroupv2` is also used in some edge cases. If the systemd hybrid mode is enabled for cgroups, then both v1 and v2 are available. Thus, Snekbox will just use whatever `use_cgroupv2` is set to. The other case is when it seems that neither v1 or v2 exist, for which a warning will be shown. If this case is reached, chances are the mount paths in the NsJail config are incorrect. The alternative is that the system does have v1 enabled, but all of the controllers are disabled.

## Tests

First, the `pipenv run test` script has been updated to install numpy into the container. There is a test that checks if numpy is importable, and it was previously failing locally because the dependency was missing.

Next, the NsJail arguments test was updated to not hard-code the indices for retrieving the arguments used in the assertion. It assumes that all NsJail args will be appended after the internal NsJail args are applied i.e. that they'll be before the `--` which separates the NsJail args from the Python args.

Finally, a test was added for #83 to prevent regressions. This involved creating a custom Gunicorn application subclass and running it in a separate process so that the test can make HTTP requests to evaluate code. This is important because it was shown that executing with NsJail directly could not reproduce the issue, but running it through Gunicorn could. The test passing is proof that v1 could have all the dynamic parent cgroup stuff removed.

## Swap

Previously, Snekbox always attempted to write to swap limit and would show a warning if it couldn't. Now, it's a bit smarter; he warning is only shown if the swap limit actually needs to be set. It considers factors like whether NsJail has a memory or swap memory limit configured and whether swap is enabled on the system. It checks for the latter by parsing `/proc/meminfo`.

It also has to check if the swap memory controller is enabled on the system. Without such check, NsJail could end up attempting to write to the file but failing because it's missing. Thus, when the controller is disabled, Snekbox will override any configured swap limit and set it to the default so that NsJail will not attempt to set a swap limit.

There are two warnings logged: one for when swap is enabled but the swap controller isn't, and another for when there's a memory limit set, but the swap is unlimited (maybe someone actually wants this, but it's likely an oversight, hence the warning).

## CI

Joe set up a self-hosted runner which uses Debian 11. This enables tests to automatically run on both a cgroupv1 and cgroupv2 system. GitHub Actions doesn't currently have a cgroupv2 runner available, so we had to host our own. The next Ubuntu release with v2 enabled by default will be 22.04, which is at least 4 months away, if not more due to waiting for GH Actions to employ it.

All steps will run on the GH Actions-hosted Ubuntu runner as they always have. However, the steps necessary to run the tests will also run on the self-hosted runner. The linting, image pushing, and deployment steps will not run on the self-hosted runner because that would be redundant.

# Known Issues

* Cgroup initialisation only happens once. However, custom NsJail arguments can be supplied which change the mounts, limits, and the version. It could re-initialise upon every call, but this _seems_ a bit expensive since it involves file I/O and may end up being redundant.
* Swap detection only happens once, but it is conceivable for it to be toggled during the lifetime of the app. Again, it could be checked upon every call, but it seems expensive since it involves file I/O; it has to search `/proc/meminfo` and for v2, even create a temp cgroup to check if the swap controller exists.
* For cgroupv1, the parent cgroups will not be deleted after Snekbox terminates. This is only significant, if at all, for when the host's cgroup namespace is used, which is the default for cgroupv1 systems with Docker.


HassanAbouelela:
Tested the PR on the following platforms:
- Ubuntu 20.04 with v1 and v2
- Ubuntu 21.10 with v1 and v2
- Debian 11 v2 (Ran out of time for v1)

Nothing seems out of the ordinary. Code eval works, resource limits work.
I was also testing Deb 10, but ran into some issues with the compose version, and didn't have time to finish testing.

jb3 (approved):

PR #129, opened 2021-12-23
Excluded snekbox/config_pb2.py from coverage

Nightfurex:


.coveragerc
```diff
@@ -3,6 +3,10 @@ branch = True
 include = snekbox/*
 omit = snekbox/api/app.py
 
+[run]
+branch = True 
+omit = snekbox/config_pb2.py
+
```
MarkKoz, line 9: You're duplicating the entire `run` section, which I don't think is correct. Instead, you should append the path to the existing `omit` on line 4. To specify multiple omit values, you can put them on separate lines, as shown in the example in the documentation: https://coverage.readthedocs.io/en/6.2/source.html#execution

MarkKoz (changes requested):

MarkKoz:
Thanks for the contribution! Would you mind editing your PR description to mention the issue being fixed? For example, "Fixes #x" where x is the issue number, would associate the PR with that issue. Thus, GitHub would automatically close the issue when the PR is merged.

.coveragerc
Nightfurex: i am not understanding how to  add multiple omit values do you have any other examples.

MarkKoz: Look at how `exclude_lines` is specified currently. Notice how each value is on a new line. The same would apply to `omit` for what you're trying to add.

Nightfurex: you mean like these omit = 
                                              snekbox/api/app.py
                                              snekbox/config_pb2.py

MarkKoz: Yes, though please indent the lines with 4 spaces, just like the values for `exclude_lines` are indented.

.coveragerc
```diff
@@ -1,7 +1,11 @@
 [run]
 branch = True
 include = snekbox/*
-omit = snekbox/api/app.py
+omit = 
+    snekbox/api/app.py
+    snekbox/config_pb2.py
+
+
```
MarkKoz, line 8: Remove these 2 extra newlines please.

MarkKoz (changes requested):

Nightfurex:
> Thanks for the contribution! Would you mind editing your PR description to mention the issue being fixed? For example, "Fixes #x" where x is the issue number, would associate the PR with that issue. Thus, GitHub would automatically close the issue when the PR is merged.

so

PR #132, opened 2021-12-26
Fix passing arguments to python -m snekbox and support Python args

### LINKED ISSUE #124: Passing arguments when running from command line fails
When running `python -m snekbox 'print("hello world!")' --time_limit 0` (documented [here](https://github.com/python-discord/snekbox/blob/main/DEVELOPING.md)) I get this error:
```python
  File "/usr/local/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/local/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/snekbox/snekbox/__main__.py", line 28, in <module>
    main()
  File "/snekbox/snekbox/__main__.py", line 23, in main
    result = NsJail().python3(args.code, *args.nsjail_args)
TypeError: NsJail.python3() takes 2 positional arguments but 4 were given
```
This is because of https://github.com/python-discord/snekbox/pull/108 changing the signature for the `NsJail.python3` function

This is the relevant line that needs changing:
https://github.com/python-discord/snekbox/blob/dc9a9e0aa9c0d56b8cbfd2e3f84f011ce0e3201d/snekbox/__main__.py#L23

The fix should be simple, just pass in the keyword argument `nsjail_args`, although do we want to try and handle setting `py_args` from the command line too? If someone suggests a suitable interface for that i'd be happy to add it, otherwise i'd be happy to just fix the passing of nsjail arguments.

MarkKoz:
Fixes #124 

NsJail arguments now properly get forwarded. There's also support for passing Python arguments. This works by separating NsJail and Python args with a  `---`. The approach discussed in the issue, which was to quote all the Python arguments, does not work because argparse treats stuff that starts with a dash as an argument rather than a string.

### DIFF
--- a/snekbox/__main__.py
+++ b/snekbox/__main__.py
@@ -44,5 +44,5 @@ def main() -> None:
         sys.exit(result.returncode)
 
 
-if __name__ == "__main__":
+if __name__ == "__main__":  # pragma: no cover
     main()
--- a/snekbox/__main__.py
+++ b/snekbox/__main__.py
@@ -20,7 +20,7 @@ def parse_args() -> argparse.Namespace:
 def main() -> None:
     """Evaluate Python code through NsJail."""
     args = parse_args()
-    result = NsJail().python3(args.code, *args.nsjail_args)
+    result = NsJail().python3(args.code, nsjail_args=args.nsjail_args)
     print(result.stdout)
 
 
--- a/snekbox/__main__.py
+++ b/snekbox/__main__.py
@@ -1,4 +1,5 @@
 import argparse
+import sys
 
 from snekbox.nsjail import NsJail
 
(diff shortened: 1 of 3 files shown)


snekbox/__main__.py
```diff
@@ -1,28 +1,48 @@
...
-    # Its actual value comes from all the unknown arguments.
+    parser.add_argument(
+        "nsjail_args", nargs="?", default=[], help="override configured NsJail options"
+    )
+    parser.add_argument(
+        "py_args", nargs="?", default=["-c"], help="arguments to pass to the Python process"
+    )
+
+    # nsjail_args and py_args are just dummies for documentation purposes.
+    # Their actual values comes from all the unknown arguments.
```
wookie184, line 23:
```suggestion
    # Their actual values come from all the unknown arguments.
```

wookie184 (commented):
Minor typo, but this seems to work nicely.

PR #133, opened 2021-12-27
CI Refactor

MarkKoz:
The workflow is broken up into separate reusable workflows. No more of the monolithic workflow file! Another consequence is that the lint workflow can run concurrently with the build workflow, since linting doesn't depend on the build.

The build uploads a tarball of the Docker image as an artefact so that other workflows can download the artefact and start a container. For tests, the coverage results from each OS in the matrix are combined into one and finally published.

I think the overall duration of the new workflow shouldn't be much longer than the current workflow, especially due to being able to run linting in parallel despite the additional dependency install taking more time.


.github/workflows/build.yaml
```diff
@@ -0,0 +1,96 @@
+on:
+  workflow_call:
+    outputs:
+      artefact:
+        description: The name of the uploaded image aretfact.
+        value: ${{ jobs.build.outputs.artefact }}
```
lemonsaurus, line 6:
```suggestion
      artifact:
        description: The name of the uploaded image artifact.
        value: ${{ jobs.build.outputs.artifact }}
```

These are more commonly called `artifacts`, [here's an ngram](https://books.google.com/ngrams/graph?content=artefact%2Cartifact&year_start=1900&year_end=2000&corpus=26&smoothing=3). It's also called artifact [by GitHub documentation](https://docs.github.com/en/actions/advanced-guides/storing-workflow-data-as-artifacts), so even though `artefact` is etymologically justifiable, I think it's probably cleaner for us to go with `artifact`.

And yes I realize this is exceptionally pedantic of me 😁 

There are a bunch of these further down, but I'll leave it up to you whether you want to make this change or not, and if you're gonna do it, probably do it with a search and replace.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,96 @@
...
+      # See https://github.com/moby/buildkit#github-actions-cache-experimental
+      #
+      # Because the cache is scoped to the branch, it will not be available
+      # on the main branch when the PR is merged. Furthermore, using this cache
+      # on main is redundant since the previous build's images are already
+      # cached on GHCR. Thus, this step is only used for PRs.
+      - name: Configure cache
+        id: cache_config
+        run: |
+          set -euo pipefail
```
lemonsaurus, line 57: this line is interesting, especially `pipefail`. What does it do and why is it relevant? I don't know if we need to document it better or not, maybe not.

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,98 @@
...
+        required: true
+        type: string
+
+jobs:
+  deploy:
+    name: Build, push, & deploy
+    runs-on: ubuntu-latest
+
+    steps:
+      # region Docker setup
```
lemonsaurus, line 17: This comment is unclear to me. Can you make it clearer?

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,98 @@
...
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Log in to GitHub Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN  }}
+      # endregion
```
lemonsaurus, line 36: oh.. I get it now. It's an IDE accordion. I don't know if that's neccessary in a short and readable workflow such as this one?

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,73 @@
...
+      - name: Show pre-commit log
+        if: always() && steps.run-pre-commit-hooks.outcome == 'failure'
+        run: cat "${PRE_COMMIT_HOME}/pre-commit.log"
+
+      # Output linting errors in the format GitHub Actions recognises for
+      # annotations.
+      - name: Run flake8
+        run: >-
+          flake8 --format "::error
+          file=%(path)s,line=%(row)d,col=%(col)d::[flake8] %(code)s: %(text)s"
```
lemonsaurus, line 73: oooh, this is fancy. I like it. 👍🏼

.github/workflows/main.yaml
```diff
@@ -0,0 +1,26 @@
...
+  build:
+    uses: ./.github/workflows/build.yaml
+  lint:
+    uses: ./.github/workflows/lint.yaml
+  test:
+    uses: ./.github/workflows/test.yaml
+    needs: build
+    with:
+      artefact: ${{ needs.build.outputs.artefact }}
+      tag: ${{ needs.build.outputs.tag }}
```
lemonsaurus, line 19: holy shit I had no idea this was legal syntax but it's _so cool_.

.github/workflows/test.yaml
```diff
@@ -0,0 +1,99 @@
...
+jobs:
+  test:
+    name: Test with coverage
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-20.04, self-hosted]
+
+    steps:
+      # region container setup
```
lemonsaurus, line 20: I generally don't love `# region`'s for any file that isn't enormous, it just feels like bloat to me. But keep them if you feel strongly about them.

lemonsaurus (approved):
Honestly, this is really _really_ good. I learned quite a bit from reading this PR. This is industry professional level CI, and it's well documented, clean, and very clever. You're using some nice modern features to keep things lean and dry.

Exceptional work. 🏆

.github/workflows/build.yaml
MarkKoz: The pipefail is actually not relevant here since there aren't any pipes. I could remove it to reduce confusion. But in general these options are really nice https://coderwall.com/p/fkfaqq/safer-bash-scripts-with-set-euxo-pipefail

.github/workflows/test.yaml
MarkKoz: I will remove them. This was left over from earlier attempts when this section was quite a bit longer.

.github/workflows/deploy.yaml
MarkKoz: Removed in e08183e

.github/workflows/test.yaml
MarkKoz: Removed in e08183e

.github/workflows/build.yaml
MarkKoz: Renamed in 6b3ec28

.github/workflows/lint.yaml
MarkKoz: We already do this in other projects, actually. I think I just copied it over from one if it wasn't already in the old snekbox CI.

PR #136, opened 2021-12-28
Store relative paths in coverage data

### LINKED ISSUE #135: Generate coverage data using relative paths
Currently, coverage data is generated with absolute paths. This means that if the coverage file is moved across machines, the source code must be in the exact same location. The current workaround is to configure Docker to mount the host's code to the same exact directory within the container. However, this doesn't work for Windows machines since the host's path is not compatible on the Linux container.

The `relative_files = true` setting can be added to `.coveragerc` to fix this.

Note that coverage seems to rely on the correct file separator being used when generating coverage reports and combining data. In practice, this means that a Windows host will be unable to generate coverage reports since the data coming from the container will have paths with forward slashes. Until this is addressed by coverage.py, the current workaround for Windows users will be to generate coverage reports within the container. This should work for `coverage report`; unsure about other formats such as HTML.

MarkKoz:
Fix #135

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -30,8 +30,6 @@ docker-compose up
 
 The container has all development dependencies. The repository on the host is mounted within the container; changes made to local files will also affect the container.
 
-Note that the compose file depends on the environment variable `PWD` being set to the current working directory. It needs it to create the aforementioned bind mount. Unix shells normally have this set already. If for some reason it is not set, it needs to be manually set. A convenient way to set it is to define it in a `.env` file which Docker Compose will automatically read.
-
 To build a normal container that can be used in production, run
 
 ```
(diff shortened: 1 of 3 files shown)


DEVELOPING.md
```diff
@@ -62,7 +60,9 @@ Alternatively, a report can be generated as HTML with
 pipenv run coverage html
 ```
 
-The HTML will output to `./htmlcov/` by default
+The HTML will output to `./htmlcov/` by default.
+
+The report cannot be generated on Windows directly due to the difference in file separators in the paths. Instead, launch a shell in the container (see below) and use `coverage report` or `coverage html` (do not invoke through Pipenv).
```
onerandomusername, line 65: Could we include the issue on the coverage docs here?

MarkKoz: I didn't want to because this problem is not quite the same as the one reported in the issue. I also don't foresee any useful information coming from that issue to warrant mentioning it.

onerandomusername: Gotcha.

onerandomusername (approved):

HassanAbouelela (approved):

Issue #138, opened 2022-03-31
Feature request: Add informational endpoint about snekbox


onerandomusername:
For those of us running our own snekbox instances, an informational endpoint would be useful. This could list the api version, the current version of snekbox, and perhaps the installed packages.

MarkKoz:
What kind of versioning would you like to see? Off the top of my head, the snekbox version could just be the commit hash and the API version could be something like semver.

onerandomusername:
That sounds good 👍

PR #140, opened 2022-05-31
Add pyproject.toml and versioning

MarkKoz:
# Summary

* Add a `pyproject.toml` using setuptools as the PEP 517 build backend
* Use setuptools's experimental PEP 621 support to define metadata in `pyproject.toml`
* Version the package using a CalVer variant which is derived from the current commit's date
* Use the version instead of the SHA to tag containers and the Sentry release
* Make Gunicorn and Sentry SDK optional dependencies (installed via extras)

The addition of versioning is progress towards #138. Originally I planned to just use the commit SHA, but I realised that it's not an adequate version since it cannot be properly constrained (except for `==`). I chose CalVer because I don't want the maintenance burden of SemVer and I don't want to have to tag releases. Deriving the version from the commit allows for the current workflow to remain: every push to master creates a new release that is deployed automatically.

# Review Process

Try installing the package (`pip install .`), importing it, and checking that the `__version__` matches the HEAD commit's date. Also run the tests with `make test` to ensure the dev container still works.


.dockerignore
```diff
@@ -2,8 +2,13 @@
 *
 
 # Make exceptions for what's needed
-!snekbox
+!.git/
```
Akarys42, line 5: This could be further filtered to speed up transfer time, but I doubt it is worth spending time on.

Dockerfile
```diff
@@ -60,18 +62,17 @@ RUN if [ -n "${DEV}" ]; \
...
 
-COPY . /snekbox
-WORKDIR /snekbox
+# ------------------------------------------------------------------------------
+FROM venv
 
-# At the end to prevent it from invalidating the layer cache.
-ARG git_sha="development"
-ENV GIT_SHA=$git_sha
+# Use a separate directory to avoid importing the source over the instaled pkg.
```
Akarys42, line 73: Small typo
```suggestion
# Use a separate directory to avoid importing the source over the installed pkg.
```

Akarys42 (approved):
Seems to work as intended! I did find one typo.

PR #143, opened 2022-06-01
Import the WSGI app via the config during tests

MarkKoz:
Relying more on Gunicorn and its config parsing will avoid discrepancies between the test and production environments. This is in response to me forgetting to call the app class in the config, but remembering to do so in the test (26c729a4d99e3f65f000030e251a4678241c599b).

Also ignore a deprecation warning from Sentry SDK's Falcon integration.

### DIFF
--- a/snekbox/utils/logging.py
+++ b/snekbox/utils/logging.py
@@ -1,6 +1,10 @@
 import logging
 import os
 import sys
+import warnings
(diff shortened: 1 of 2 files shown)


HassanAbouelela (approved):
Tested, everything looks good

PR #145, opened 2022-06-07
Make the output size limit customisable

### LINKED ISSUE #104: Make the Output Size Limit Customisable
Currently, the limit is hard-capped at 1 MB here

https://github.com/python-discord/snekbox/blob/93f99fb909a60b2bcc12c4967adf767c0a82c804/snekbox/nsjail.py#L30-L31

These should read from environment variables to enable users to customise the values when launching the container.

MarkKoz:
Resolves #104 

The `NsJail` object can now take more arguments. These can be specified by modifying the WSGI app URI in the Gunicorn config. See the updated README.md for more info.

Also refactored some minor things and fixed a mistake in black's config.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ All environment variables have defaults and are therefore not required to be set
 
 Name | Description
 ---- | -----------
-`DEBUG` | Enable debug logging if set to a non-empty value.
+`SNEKBOX_DEBUG` | Enable debug logging if set to a non-empty value.
 `SNEKBOX_SENTRY_DSN` | [Data Source Name] for Sentry. Sentry is disabled if left unset.
 
 ## Third-party Packages
--- a/README.md
+++ b/README.md
@@ -66,7 +66,9 @@ NsJail is configured through [`snekbox.cfg`]. It contains the exact values for t
 
 ### Gunicorn
 
-[Gunicorn settings] can be found in [`gunicorn.conf.py`]. In the default configuration, the worker count and the bind address are likely the only things of any interest. Since it uses the default synchronous workers, the [worker count] effectively determines how many concurrent code evaluations can be performed.
+[Gunicorn settings] can be found in [`gunicorn.conf.py`]. In the default configuration, the worker count, the bind address, and the WSGI app URI are likely the only things of any interest. Since it uses the default synchronous workers, the [worker count] effectively determines how many concurrent code evaluations can be performed.
+
+`wsgi_app` can be given arguments which are forwarded to the `NsJail` object. For example, `wsgi_app = "snekbox:SnekAPI(max_output_size=2_000_000, read_chunk_size=20_000)"`.
 
 ### Environment Variables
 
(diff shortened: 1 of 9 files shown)


snekbox/nsjail.py
```diff
@@ -79,10 +80,6 @@ def _parse_log(log_lines: Iterable[str]) -> None:
                 continue
 
             msg = match["msg"]
-            if not DEBUG and any(msg.startswith(s) for s in LOG_BLACKLIST):
-                # Skip blacklisted messages if not debugging.
-                continue
-
```
Bluenix2, line 85: Why was this removed? I don't think it relates to the other changes made here?

tests/test_nsjail.py
```diff
@@ -280,3 +282,38 @@ def test_py_args(self):
...
+        cls.addClassCleanup(cls._temp_dir.cleanup)
+
+        cls.nsjail_path = shutil.copy2("/usr/sbin/nsjail", cls._temp_dir.name)
+        cls.config_path = shutil.copy2("./config/snekbox.cfg", cls._temp_dir.name)
+        cls.max_output_size = 1_234_567
+        cls.read_chunk_size = 12_345
+
+        cls.nsjail = NsJail(
+            cls.nsjail_path, cls.config_path, cls.max_output_size, cls.read_chunk_size
+        )
```
Bluenix2, line 300: Should this really be shared across tests? I think this should be placed in `def setUp(self):` instead.

README.md
```diff
@@ -125,3 +123,4 @@ See [CONTRIBUTING.md](.github/CONTRIBUTING.md).
 [sentry release]: https://docs.sentry.io/platforms/python/configuration/releases/
 [data source name]: https://docs.sentry.io/product/sentry-basics/dsn-explainer/
 [GitHub Container Registry]: https://github.com/orgs/python-discord/packages/container/package/snekbox
+[`NsJail`]: snekbox/nsjail.py
```
Bluenix2, line 126: I don't think this linking worked correctly. I was not able to press `NsJail` and get linked to this file.

Bluenix2 (changes requested):
Looks good; I have a few comments though.

On top of that, please edit this line as well in the README since the output size can now be configured:
> The output returned by snekbox is truncated at around 1 MB.

tests/test_nsjail.py
MarkKoz: The instance is designed to be re-used but I suppose it is safer to isolate test dependencies. 308385c

README.md
MarkKoz: [55270d2](https://github.com/python-discord/snekbox/pull/145/commits/55270d29dac5c88f8cd70324b0febb6c792b05e4)

MarkKoz: Sorry, I made the change but forgot to copy it over to the file git was tracking.

MarkKoz: c9bfff8

Bluenix2 (approved):
Looks great! Thank you

PR #147, opened 2022-07-04
Create v1 cgroups as needed

### LINKED ISSUE #101: Create all (needed) parent cgroups
Currently, only CPU and memory parent cgroups are created. While this works fine for the default config, it doesn't account for custom configs which may use additional cgroups. It should either create all possible parents or try to read the config and determine which cgroups are needed. If the latter is possible and not complicated to implement, then it would be preferable to the former.

Special cases, similar to disabling swap for the memory cgroup, should be investigated when creating the other parent cgroups.

MarkKoz:
Resolves #101

NsJail puts subprocess cgroups under parent cgroups. However, it does not create the parents automatically. Thus, snekbox has to do this. Previously, for v1, only the `memory` and `pids` cgroup parents were being created. Now, all four controllers supported by NsJail are also supported by snekbox.

It looks at the NsJail config to determine if a cgroup parent needs to be created. Basically, if a value is non-default, then NsJail will create a cgroup and thus will rely on a parent existing.

For the test, I test each controller separately. Each controller has several test cases for the various combinations of config values (default vs explicit). However, I don't think the mount value actually has any effect since it's equivalent to the default. I can't use a non-default value because I don't think it's easy to change where the cgroup fs is mounted if it's already mounted.

### DIFF
--- a/tests/test_nsjail.py
+++ b/tests/test_nsjail.py
@@ -336,8 +336,8 @@ class NsJailCgroupTests(unittest.TestCase):
                 rw: false
             }
             exec_bin {
-                path: "/usr/local/bin/python"
-                arg: "-Squ"
+                path: "/bin/su"
+                arg: ""
             }
             """
         ).strip()
@@ -383,4 +383,4 @@ class NsJailCgroupTests(unittest.TestCase):
 
                         result = nsjail.python3("")
 
-                        self.assertEqual(result.returncode, 0)
+                        self.assertNotEqual(result.returncode, 255)
--- a/tests/test_nsjail.py
+++ b/tests/test_nsjail.py
@@ -5,6 +5,8 @@ import sys
 import tempfile
 import unittest
 import unittest.mock
+from itertools import product
+from pathlib import Path
 from textwrap import dedent
 
 from snekbox.nsjail import NsJail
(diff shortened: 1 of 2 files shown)


tests/test_nsjail.py
```diff
@@ -316,3 +318,69 @@ def test_init_args(self):
...
+
+        config_base = dedent(
+            """
+            mode: ONCE
+            mount {
+                src: "/"
+                dst: "/"
+                is_bind: true
+                rw: false
+            }
```
HassanAbouelela, line 337: This binding only works if the installed python version matches what you're expecting (`/usr/local/bin/python`), but will cause the test to fail if the user defines a non-default mount or exec_bin. Could we load the mounts and exec_bin from the actual config into here?

HassanAbouelela (changes requested):
Everything works as expected, and I've tested that the limits do in fact get applied correctly. The PR looks good to me, with the exception of my comment.

tests/test_nsjail.py
HassanAbouelela: Something like this seems to do the trick
```py
        user_config = NsJail._read_config("./config/snekbox.cfg")
        mounts = "\n".join((f"mount {{{mount}}}" for mount in user_config.mount))
        config_base = dedent(
            f"""
            {mounts}\n
            exec_bin {{{user_config.exec_bin}}}
            """
        ).strip()
```

[Patch included](https://github.com/python-discord/snekbox/files/9072325/Use_User_Config_In_Tests.txt)

MarkKoz: The goal of this test is to ensure NsJail does not crash when each of the cgroups is configured. If the parent cgroup is not created properly, then NsJail will crash. Which process actually runs is of no concern for this test; I just had to specify _something_ so NsJail can run. The tests run in the image, and we know the image contains Python, so using Python is a safe bet.

HassanAbouelela: This is only true for users using the provided docker image, but on local installations, or on modified versions (which is my use-case), it will error. If we don't care about the specific command being python, let's use something else that's likely to exist in all environments. Perhaps something like `ls` could be a good choice.

HassanAbouelela (approved):

PR #153, opened 2022-10-31
Bump To 3.11

### LINKED ISSUE #149: Use modern Docker Compose command
The Compose file uses some features of the spec only supported by Docker Compose v2, such as `pull_policy`. The contribution documentation and the Makefile currently refer to the v1 command, `docker-compose`. Under some circumstances, this can be aliased to v2, but it isn't guaranteed to always be the case. All references should be updated to use the v2 command, `docker compose`. An explicit mention of v2 being required for the command would be helpful as well.

### LINKED ISSUE #152: python 3.11 support
When might we expect snekbox to have a 3.11 build?

HassanAbouelela:
This PR bumps the snekbox version to the latest stable python version. Closes #152, closes #149
- [x] Bump version in dockerfile
- [x] Bump version in documentation
- [x] Relock dependency versions
- [x] ~~Possibly look into bumping dependency versions?~~
- [x] Upgrade and test runtime packages

I'll upgrade where possible to avoid the image having to compile a bunch of packages itself, to speed up builds, however we might have to revisit in the future as I imagine a lot of packages don't have wheels yet.


HassanAbouelela:
I might also go ahead and implement #149 since it's a relatively small change. I've already been using compose v2 and everything works as expected of course.

.pre-commit-config.yaml
```diff
@@ -20,7 +20,7 @@ repos:
     rev: 22.3.0
     hooks:
       - id: black
-        language_version: "3.10"
```
MarkKoz, line 23: You missed https://github.com/python-discord/snekbox/blob/f2f1193ffdf7efa282f09555a3dc1501bfdbc5dd/pyproject.toml#L67

Can you check if these settings are redundant? If so, let's remove it from pre-commit-config.

HassanAbouelela: Will test and report back

MarkKoz:
> Possibly look into bumping dependency versions?

The dependencies only have lower bounds. I think they would only need to be updated if there is a package whose metadata indicates it works with 3.10 and higher, but in practice it doesn't actually support 3.11.

HassanAbouelela:
I think you're right on the package dep front.

HassanAbouelela:
Finished upgrading and testing runtime packages. All went fine. The following packages needed to build wheels:
> forbiddenfruit, pendulum, yarl, multidict

We might want to upgrade them in the future, but overall very little building necessary

.pre-commit-config.yaml
HassanAbouelela: As far as I could tell, it did manage to successfully read and use the target-version from here. That said, the target-version flag has very little effect on the output in general, so being certain is relatievly difficult.

MarkKoz (approved):
Thanks!

PR #155, opened 2022-11-05
stop using the deprecated set-output command

onerandomusername:
Looking at the [last action run](https://github.com/python-discord/snekbox/actions/runs/3399634090), there were over 50 warnings from github about using deprecated actions and warnings
This solves the simplest of them by replacing all of the `::set-output` commands with the new syntax.

See [the changelog](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) for more information.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -30,7 +30,7 @@ jobs:
         run: |
           set -eu
           version=$(python scripts/version.py)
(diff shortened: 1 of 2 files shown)


ChrisLovering (approved):

ChrisLovering:
Thanks!

PR #156, opened 2022-11-05
upgrade all gha packages to their latest node16 versions

onerandomusername:
got another one here 😄 

I've checked the [changelogs](https://github.com/actions/checkout/releases) and upgraded this on my own projects, this seems to have the single change of changing from node 12 to node 16 which solves another bunch of warnings.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -20,7 +20,7 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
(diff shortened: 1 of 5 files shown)


onerandomusername:
I've changed this to upgrade *every* gha package that has an update which should solve most if not all of the deprecation warnings

ChrisLovering (approved):

onerandomusername:
I missed an actions/checkout@v2 in test.yaml. Not a big deal, as azure/setup-kubectl will need a bump at some point once they release a new version that doesn't use the set-ouput command

Issue #103, opened 2021-04-08
Support Output of Raw Bytes from the Subprocess


MarkKoz:
Currently, all output is interpreted as Unicode text. Supporting byte output would allow a much more diverse range of things that could be returned e.g. images and sound. I think it makes sense to maintain a single endpoint and simply add a parameter to change from text to byte mode, but I'm still open to discussion on making it a separate endpoint.

Since the API uses JSON, the bytes would have to be base64-encoded, unless someone has a better alternative (which could perhaps also be a good argument for making it a separate endpoint).

Raising the output limit from 1 MB should be considered for byte output, if not all output. 8 MB sounds sensible and makes sense in the context of Discord attachments (even though Python Discord does have a higher limit thanks to boosts). It'd be useful to have someone from @python-discord/devops consult on an appropriate limit.

ChrisLovering:
I guess we could leverage the use of an `Accept` header in the post call to determine what to send back, so that it's backwards compatible, but also gives the option to accept an octet stream, or base64 encoded bytes.

MarkKoz:
@ChrisLovering That is an interesting idea, but the response also needs to include the return code of the process. I don't see how that could work.

MarkKoz:
@python-discord/devops Can you advise on a reasonable output limit? Compressing responses is also an option (https://github.com/willnewton/falcon-compression), but not sure if it's necessary. What do you think?

ChrisLovering:
Anywhere 5-10MB seems reasonable to me

MarkKoz:
And what of compression?

HassanAbouelela:
We have more CPU resources than we can use so if you think it'll prove useful, compression sounds fine.

PR #159, opened 2022-11-15
File system and Binary file sending

### LINKED ISSUE #114: Support File I/O
Several users have expressed a desire to interact with files with their executed code. I believe such feature is in scope and that it would be a great feature. Supporting it would require a way to mount an isolated, size-capped r/w directory for each NsJail instance. Implementation details need to be investigated to determine the feasability of this feature.

ionite34:
Closes #114, #103

## Permission specifications
1. We will create a container directory at `/snekbox/memfs`, which will be mounted as a tmpfs memory directory. The `memfs` folder is execute-only for user code with no read/write permissions.
2. Each snekbox run instance will have a new mounted tmpfs on the docker system created with a unique ID, a subfolder of this (`/home`) is then mounted as R/W folders onto the NSJail instance.
3. No instance should be able to read other instance filesystem mounts as they are not mounted to NSJail. 
```yaml
# Docker host
|- /memfs/ab043b4e-5955-4734-b8e2-998c0f6c7131/home/main.py

# Visible to instance user
|- /home/main.py
```
4. This UUID directory is read/execute only for the user code
5. There is also a `home` folder, which has read/write permission for nsjail, and is mounted into nsjail as a R/W folder at `/home`
6. The `$HOME` env var has also been set to point to this mounted folder in the nsjail. (Required for libraries like matplotlib to work)

## Usage
The user has a cwd in `/home` which they can read/write in. In this directory there will be a `output` folder. Any files written within this folder will be uploaded.

The file system per instance is default to be limited at 48MiB, can be changed in the cfg file.

### File system
- [x] 1. Implement a unique tempdir generator backed by `dev/shm`
- [x] 2. Change runtime config to mount an unique tempdir to each snekbox execution
- [x] 3. Make this the working directory for the python file
- [x] 4. Use a written python file for better compatibility with file usage (`__file__`) and other usages such as `inspect.getsource`

### Sending binary files via API
- [x] 1. Scan for created files after snekbox execution completes
- [x] 2. Serialize binary file into JSON, with data and name (possible base64 encoding and / or compression)
- [x] 3. Add API route
- [x] 4. Add a user-code side method for files to be marked sent, or some standard (like naming the file "out.*")
- [x] 5. Formalize API route specifications and docs

Compatible mode with input
```py
{
    "input": "[i for i in range(1000)]",
    "args": ["-m", "timeit"]
}
```
New args without input
```py
{
    "args": ["-m", "timeit", "[i for i in range(1000)]"]
}
```
Sending files to be run
```py
{
    "args": ["main.py"],
    "files": [
        {
            "name": "main.py",
            "content": "print(1)"
        }
    ]
}
```

## API response additions

(Previous)
```py
{
	"stdout": "Hello\n",
	"returncode": 0
}
```

(New)
New dictionary key `files` of type `list[dict]`, holding dictionaries representing attachments
Sub-dictionaries will have type `dict[str, str | int]`
Attachment dict keys:
- "name" (str)
- "size" (int) Content size in bytes (before compression)
- "content" (str) Base64 encoded bytes
```py
{
	"stdout": "Hello\n",
	"returncode": 0,
	"files": [
		{
			"name": "output.png",
			"size": 57344,
			"content": "SGVsbG8...=",
		},
	]
}
```

## Demo Images

image
image

image

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,28 +70,22 @@ NsJail is configured through [`snekbox.cfg`]. It contains the exact values for t
 
 ### Memory File System
 
-On each execution, the host will mount an instance-specific `tmpfs` directory,
-this is used as a limited read-write folder for the sandboxed code. There is no
-access of any kind to other files or directories on the host system or container,
-as only this new directory is mounted to NSJail.
+On each execution, the host will mount an instance-specific `tmpfs` drive, this is used as a limited read-write folder for the sandboxed code. There is no access to other files or directories on the host container beyond the other read-only mounted system folders. Instance file systems are isolated, it is not possible for sandboxed code to access another instance's writeable directory.
 
-The following options for the memory file system are configurable as options in
-[gunicorn.conf.py](config/gunicorn.conf.py)
+The following options for the memory file system are configurable as options in [gunicorn.conf.py](config/gunicorn.conf.py)
 
 * `memfs_instance_size` Size in bytes for the capacity of each instance file system.
 * `files_limit` Maximum number of valid output files to parse.
 * `files_timeout` Maximum time in seconds for output file parsing and encoding.
 * `files_pattern` Glob pattern to match files within `output`.
 
-The sandboxed code execution will start with a working directory of `home`, and
-a visible folder `output`. The user has read/write access to any path under `home`.
+The sandboxed code execution will start with a working directory of `home`, and a visible folder `output`. The user has read/write access to any path under `home`.
 ```
 /home
  |- output
 ```
 
-Files written to the `output` subfolder will be parsed and returned as
-a list of `FileAttachment` objects in `EvalResult.files` from the `python3` function.
+Files written to the `output` subfolder will be parsed and returned as a list of `FileAttachment` objects in `EvalResult.files` from the `python3` function.
 
 To send files to snekbox, it can be included as the `files` parameter of `python3`.
 
--- a/README.md
+++ b/README.md
@@ -7,6 +7,9 @@
 
 Python sandbox runners for executing code in isolation aka snekbox.
 
+Supports a memory [virtual read/write file system](#virtual-file-system) within the sandbox,
+allowing text or binary files to be sent and returned.
+
 A client sends Python code to a snekbox, the snekbox executes the code, and finally the results of the execution are returned to the client.
 
 ```mermaid
@@ -60,9 +63,41 @@ The main features of the default configuration are:
 * Memory limit
 * Process count limit
 * No networking
-* Restricted, read-only filesystem
+* Restricted, read-only system filesystem
+* Memory-based virtual read-write filesystem mounted as cwd at `/home`
 
 NsJail is configured through [`snekbox.cfg`]. It contains the exact values for the items listed above. The configuration format is defined by a [protobuf file][7] which can be referred to for documentation. The command-line options of NsJail can also serve as documentation since they closely follow the config file format.
+
+### Memory File System
+
+On each execution, the host will mount an instance-specific `tmpfs` directory,
+this is used as a limited read-write folder for the sandboxed code. There is no
+access of any kind to other files or directories on the host system or container,
+as only this new directory is mounted to NSJail.
+
+The following options for the memory file system are configurable as options in
+[gunicorn.conf.py](config/gunicorn.conf.py)
+
+* `memfs_instance_size` Size in bytes for the capacity of each instance file system.
+* `files_limit` Maximum number of valid output files to parse.
+* `files_timeout` Maximum time in seconds for output file parsing and encoding.
+* `files_pattern` Glob pattern to match files within `output`.
+
+The sandboxed code execution will start with a working directory of `home`, and
+a visible folder `output`. The user has read/write access to any path under `home`.
+```
+/home
+ |- output
+```
+
+Files written to the `output` subfolder will be parsed and returned as
+a list of `FileAttachment` objects in `EvalResult.files` from the `python3` function.
+
+To send files to snekbox, it can be included as the `files` parameter of `python3`.
+
+Within the `/eval` route, files are attached or returned under the `files` key.
+
+> See [api/resources/eval.py](snekbox/api/resources/eval.py) for additional API schema information.
 
 ### Gunicorn
 
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ NsJail is configured through [`snekbox.cfg`]. It contains the exact values for t
 
 ### Memory File System
 
-On each execution, the host will mount an instance-specific `tmpfs` drive, this is used as a limited read-write folder for the sandboxed code. There is no access to other files or directories on the host container beyond the other read-only mounted system folders. Instance file systems are isolated, it is not possible for sandboxed code to access another instance's writeable directory.
+On each execution, the host will mount an instance-specific `tmpfs` drive, this is used as a limited read-write folder for the sandboxed code. There is no access to other files or directories on the host container beyond the other read-only mounted system folders. Instance file systems are isolated; it is not possible for sandboxed code to access another instance's writeable directory.
 
 The following options for the memory file system are configurable as options in [gunicorn.conf.py](config/gunicorn.conf.py)
 
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ The main features of the default configuration are:
 * Process count limit
 * No networking
 * Restricted, read-only system filesystem
-* Memory-based virtual read-write filesystem mounted as cwd at `/home`
+* Memory-based read-write filesystem mounted as working directory `/home`
 
 NsJail is configured through [`snekbox.cfg`]. It contains the exact values for the items listed above. The configuration format is defined by a [protobuf file][7] which can be referred to for documentation. The command-line options of NsJail can also serve as documentation since they closely follow the config file format.
 
@@ -75,15 +75,13 @@ On each execution, the host will mount an instance-specific `tmpfs` drive, this 
 The following options for the memory file system are configurable as options in [gunicorn.conf.py](config/gunicorn.conf.py)
 
 * `memfs_instance_size` Size in bytes for the capacity of each instance file system.
+* `memfs_home` Path to the home directory within the instance file system.
+* `memfs_output` Path to the output directory within the instance file system.
 * `files_limit` Maximum number of valid output files to parse.
 * `files_timeout` Maximum time in seconds for output file parsing and encoding.
 * `files_pattern` Glob pattern to match files within `output`.
 
-The sandboxed code execution will start with a writeable working directory of `home`. Any files written within the subfolder `output` will be parsed for output.
-```
-/home
- |- output
-```
+The sandboxed code execution will start with a writeable working directory of `home`. By default, the output folder is also `home`. New files, and uploaded files with a newer last modified time, will be uploaded on completion.
 
 ### Gunicorn
 
--- a/README.md
+++ b/README.md
@@ -91,8 +91,6 @@ To send files to snekbox, it can be included as the `files` parameter of `python
 
 Within the `/eval` route, files are attached or returned under the `files` key.
 
-> See [api/resources/eval.py](snekbox/api/resources/eval.py) for additional API schema information.
-
 ### Gunicorn
 
 [Gunicorn settings] can be found in [`gunicorn.conf.py`]. In the default configuration, the worker count, the bind address, and the WSGI app URI are likely the only things of any interest. Since it uses the default synchronous workers, the [worker count] effectively determines how many concurrent code evaluations can be performed.
--- a/README.md
+++ b/README.md
@@ -79,17 +79,11 @@ The following options for the memory file system are configurable as options in 
 * `files_timeout` Maximum time in seconds for output file parsing and encoding.
 * `files_pattern` Glob pattern to match files within `output`.
 
-The sandboxed code execution will start with a working directory of `home`, and a visible folder `output`. The user has read/write access to any path under `home`.
+The sandboxed code execution will start with a writeable working directory of `home`. Any files written within the subfolder `output` will be parsed for output.
 ```
 /home
  |- output
 ```
-
-Files written to the `output` subfolder will be parsed and returned as a list of `FileAttachment` objects in `EvalResult.files` from the `python3` function.
-
-To send files to snekbox, it can be included as the `files` parameter of `python3`.
-
-Within the `/eval` route, files are attached or returned under the `files` key.
 
 ### Gunicorn
 
--- a/config/snekbox.cfg
+++ b/config/snekbox.cfg
@@ -21,6 +21,7 @@ envar: "HOME=home"
 keep_caps: false
 
 rlimit_as: 700
+rlimit_fsize_type: INF
 
 clone_newnet: true
 clone_newuser: true
--- a/config/snekbox.cfg
+++ b/config/snekbox.cfg
@@ -117,5 +117,5 @@ iface_no_lo: true
 
 exec_bin {
     path: "/usr/local/bin/python"
-    arg: "-Squ"
+    arg: "-BSqu"
 }
--- a/snekbox/api/resources/eval.py
+++ b/snekbox/api/resources/eval.py
@@ -27,6 +27,7 @@ class EvalResource:
     REQ_SCHEMA = {
         "type": "object",
         "properties": {
+            "input": {"type": "string"},
             "args": {"type": "array", "items": {"type": "string"}},
             "files": {
                 "type": "array",
@@ -44,7 +45,10 @@ class EvalResource:
                 },
             },
         },
-        "required": ["args"],
+        "anyOf": [
+            {"required": ["input"]},
+            {"required": ["args"]},
+        ],
     }
 
     def __init__(self, nsjail: NsJail):
@@ -57,6 +61,11 @@ class EvalResource:
 
         A list of arguments for the Python subprocess can be specified as `args`.
 
+        If `input` is specified, it will be appended as the last argument to `args`,
+        and `args` will have a default argument of `"-c"`.
+
+        Either `input` or `args` must be specified.
+
         The return codes mostly resemble those of a Unix shell. Some noteworthy cases:
 
         - None
@@ -67,6 +76,10 @@ class EvalResource:
             NsJail encountered a fatal error
 
         Request body:
+
+        >>> {
+        ...    "input": "print('Hello')"
+        ... }
 
         >>> {
         ...    "args": ["-c", "print('Hello')"]
@@ -105,10 +118,14 @@ class EvalResource:
         - 415
             Unsupported content type; only application/JSON is supported
         """
+        body: dict[str, str | list[str] | list[dict[str, str]]] = req.media
+        # If `input` is supplied, default `args` to `-c`
+        if "input" in body:
+            body.setdefault("args", ["-c"])
         try:
             result = self.nsjail.python3(
-                py_args=req.media["args"],
-                files=[FileAttachment.from_dict(file) for file in req.media.get("files", [])],
+                py_args=body["args"],
+                files=[FileAttachment.from_dict(file) for file in body.get("files", [])],
             )
         except ParsingError as e:
             raise falcon.HTTPBadRequest(title="Request file is invalid", description=str(e))
--- a/snekbox/api/resources/eval.py
+++ b/snekbox/api/resources/eval.py
@@ -9,7 +9,7 @@ from snekbox.nsjail import NsJail
 
 __all__ = ("EvalResource",)
 
-from snekbox.snekio import FileAttachment, IllegalPathError
+from snekbox.snekio import FileAttachment, ParsingError
 
 log = logging.getLogger(__name__)
 
@@ -35,7 +35,7 @@ class EvalResource:
                     "properties": {
                         "path": {
                             "type": "string",
-                            # Disallow single forward slashes, absolute paths, and null bytes
+                            # Disallow single absolute paths, and null bytes
                             "pattern": r"^[^/\\0].*",
                         },
                         "content": {"type": "string"},
@@ -110,10 +110,8 @@ class EvalResource:
                 py_args=req.media["args"],
                 files=[FileAttachment.from_dict(file) for file in req.media.get("files", [])],
             )
-        except IllegalPathError as e:
-            raise falcon.HTTPBadRequest(
-                title="Request file path failed validation", description=str(e)
-            )
+        except ParsingError as e:
+            raise falcon.HTTPBadRequest(title="Request file is invalid", description=str(e))
         except Exception:
             log.exception("An exception occurred while trying to process the request")
             raise falcon.HTTPInternalServerError
--- a/snekbox/api/resources/eval.py
+++ b/snekbox/api/resources/eval.py
@@ -122,6 +122,7 @@ class EvalResource:
         # If `input` is supplied, default `args` to `-c`
         if "input" in body:
             body.setdefault("args", ["-c"])
+            body["args"].append(body["input"])
         try:
             result = self.nsjail.python3(
                 py_args=body["args"],
--- a/snekbox/api/resources/eval.py
+++ b/snekbox/api/resources/eval.py
@@ -117,5 +117,5 @@ class EvalResource:
         resp.media = {
             "stdout": result.stdout,
             "returncode": result.returncode,
-            "files": [f.to_dict() for f in result.files],
+            "files": [f.json for f in result.files],
         }
--- a/snekbox/api/resources/eval.py
+++ b/snekbox/api/resources/eval.py
@@ -117,5 +117,5 @@ class EvalResource:
         resp.media = {
             "stdout": result.stdout,
             "returncode": result.returncode,
-            "files": [f.json for f in result.files],
+            "files": [f.as_dict for f in result.files],
         }
--- a/snekbox/api/resources/eval.py
+++ b/snekbox/api/resources/eval.py
@@ -33,7 +33,11 @@ class EvalResource:
                 "items": {
                     "type": "object",
                     "properties": {
-                        "path": {"type": "string"},
+                        "path": {
+                            "type": "string",
+                            # Disallow single forward slashes, absolute paths, and null bytes
+                            "pattern": r"^[^/\\0].*",
+                        },
                         "content": {"type": "string"},
                     },
                     "required": ["path"],
--- a/snekbox/api/resources/eval.py
+++ b/snekbox/api/resources/eval.py
@@ -36,8 +36,8 @@ class EvalResource:
                     "properties": {
                         "path": {
                             "type": "string",
-                            # Disallow single absolute paths, and null bytes
-                            "pattern": r"^[^/\\0].*",
+                            # Disallow starting with / or containing \0 anywhere
+                            "pattern": r"^(?!/)(?!.*\\0).*$",
                         },
                         "content": {"type": "string"},
                     },
--- a/snekbox/filesystem.py
+++ b/snekbox/filesystem.py
@@ -23,10 +23,10 @@ libc.umount2.argtypes = (ctypes.c_char_p, ctypes.c_int)
 class Size(IntEnum):
     """Size multipliers for bytes."""
 
-    KB = 1024
-    MB = 1024**2
-    GB = 1024**3
-    TB = 1024**4
+    KiB = 1024
+    MiB = 1024**2
+    GiB = 1024**3
+    TiB = 1024**4
 
 
 class UnmountFlags(IntEnum):
--- a/snekbox/filesystem.py
+++ b/snekbox/filesystem.py
@@ -20,13 +20,13 @@ libc.mount.argtypes = (
 libc.umount2.argtypes = (ctypes.c_char_p, ctypes.c_int)
 
 
-class Size(int):
-    """Size in bytes."""
+class Size(IntEnum):
+    """Size multipliers for bytes."""
 
-    @classmethod
-    def from_mb(cls, mb: int) -> Size:
-        """Create a Size from mebibytes."""
-        return cls(mb * 1024 * 1024)
+    KB = 1024
+    MB = 1024**2
+    GB = 1024**3
+    TB = 1024**4
 
 
 class UnmountFlags(IntEnum):
--- a/snekbox/filesystem.py
+++ b/snekbox/filesystem.py
@@ -1,3 +1,4 @@
+"""Mounts and unmounts filesystems."""
 from __future__ import annotations
 
 import ctypes
@@ -48,6 +49,9 @@ def mount(source: Path | str, target: Path | str, fs: str, **options: str | int)
         target: Target directory.
         fs: Filesystem type.
         **options: Mount options.
+
+    Raises:
+        OSError: On any mount error.
     """
     if Path(target).is_mount():
         raise OSError(f"{target} is already a mount point")
@@ -71,6 +75,9 @@ def unmount(target: Path | str, flags: UnmountFlags | int = UnmountFlags.MNT_DET
     Args:
         target: Target directory.
         flags: Unmount flags.
+
+    Raises:
+        OSError: On any unmount error.
     """
     if not Path(target).is_mount():
         raise OSError(f"{target} is not a mount point")
--- a/snekbox/libmount.py
+++ b/snekbox/libmount.py
@@ -49,6 +49,9 @@ def mount(source: Path | str, target: Path | str, fs: str, **options: str | int)
         fs: Filesystem type.
         **options: Mount options.
     """
+    if Path(target).is_mount():
+        raise OSError(f"{target} is already a mount point")
+
     kwargs = " ".join(f"{key}={value}" for key, value in options.items())
 
     result: int = libc.mount(
@@ -69,6 +72,9 @@ def unmount(target: Path | str, flags: UnmountFlags | int = UnmountFlags.MNT_DET
         target: Target directory.
         flags: Unmount flags.
     """
+    if not Path(target).is_mount():
+        raise OSError(f"{target} is not a mount point")
+
     result: int = libc.umount2(str(target).encode(), int(flags))
     if result < 0:
         errno = ctypes.get_errno()
--- a/snekbox/memfs.py
+++ b/snekbox/memfs.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 import logging
 import warnings
 import weakref
-from collections.abc import Generator, Sequence
+from collections.abc import Generator
 from contextlib import suppress
 from pathlib import Path
 from types import TracebackType
@@ -121,7 +121,10 @@ class MemFS:
         return folder
 
     def files(
-        self, limit: int, pattern: str = "**/*", exclude_files: Sequence[str] = ()
+        self,
+        limit: int,
+        pattern: str = "**/*",
+        exclude_files: dict[Path, float] | None = None,
     ) -> Generator[FileAttachment, None, None]:
         """
         Yields FileAttachments for files found in the output directory.
@@ -129,13 +132,18 @@ class MemFS:
         Args:
             limit: The maximum number of files to parse.
             pattern: The glob pattern to match files against.
-            exclude_files: Files (relative to home directory) to exclude.
+            exclude_files: A dict of Paths and last modified times.
+                Files will be excluded if their last modified time
+                is equal to the provided value.
         """
         count = 0
         for file in self.output.rglob(pattern):
-            if str(file.relative_to(self.home)) in exclude_files:
-                log.info(f"Skipping excluded file {file!s}")
-                continue
+            if exclude_files and (orig_time := exclude_files.get(file)):
+                new_time = file.stat().st_mtime
+                log.info(f"Checking {file.name} ({orig_time=}, {new_time=})")
+                if file.stat().st_mtime == orig_time:
+                    log.info(f"Skipping {file.name} as it has not been modified")
+                    continue
             if count > limit:
                 log.info(f"Max attachments {limit} reached, skipping remaining files")
                 break
@@ -149,7 +157,7 @@ class MemFS:
         limit: int,
         pattern: str,
         preload_dict: bool = False,
-        exclude_files: Sequence[str] = (),
+        exclude_files: dict[Path, float] | None = None,
     ) -> list[FileAttachment]:
         """
         Return a sorted list of file paths within the output directory.
@@ -158,7 +166,9 @@ class MemFS:
             limit: The maximum number of files to parse.
             pattern: The glob pattern to match files against.
             preload_dict: Whether to preload as_dict property data.
-            exclude_files: Files (relative to home directory) to exclude.
+            exclude_files: A dict of Paths and last modified times.
+                Files will be excluded if their last modified time
+                is equal to the provided value.
         Returns:
             List of FileAttachments sorted lexically by path name.
         """
--- a/snekbox/memfs.py
+++ b/snekbox/memfs.py
@@ -21,6 +21,7 @@ def parse_files(
     fs: MemFS,
     files_limit: int,
     files_pattern: str,
+    preload_json: bool = False,
 ) -> list[FileAttachment]:
     """
     Parse files in a MemFS.
@@ -29,11 +30,16 @@ def parse_files(
         fs: The MemFS to parse.
         files_limit: The maximum number of files to parse.
         files_pattern: The glob pattern to match files against.
+        preload_json: Whether to preload and cache JSON data.
 
     Returns:
         List of FileAttachments sorted lexically by path name.
     """
-    return sorted(fs.attachments(files_limit, files_pattern), key=lambda file: file.path)
+    res = sorted(fs.attachments(files_limit, files_pattern), key=lambda file: file.path)
+    if preload_json:
+        for file in res:
+            _ = file.json
+    return res
 
 
 class MemFS:
--- a/snekbox/memfs.py
+++ b/snekbox/memfs.py
@@ -4,9 +4,8 @@ from __future__ import annotations
 import logging
 import warnings
 import weakref
-from collections.abc import Generator, Sequence
+from collections.abc import Generator
 from contextlib import suppress
-from fnmatch import fnmatch
 from pathlib import Path
 from types import TracebackType
 from typing import Type
@@ -125,7 +124,6 @@ class MemFS:
         self,
         limit: int,
         pattern: str = "**/*",
-        ignores: Sequence[str] = (),
         exclude_files: dict[Path, float] | None = None,
     ) -> Generator[FileAttachment, None, None]:
         """
@@ -134,20 +132,12 @@ class MemFS:
         Args:
             limit: The maximum number of files to parse.
             pattern: The glob pattern to match files against.
-            ignores: A sequence of fnmatch patterns to ignore.
             exclude_files: A dict of Paths and last modified times.
                 Files will be excluded if their last modified time
                 is equal to the provided value.
         """
         count = 0
         for file in self.output.rglob(pattern):
-            if any(
-                fnmatch(str(file.relative_to(self.home)), match_pattern := ignore_pattern)
-                for ignore_pattern in ignores
-            ):
-                log.info(f"Ignoring {file.name!r} as it matches {match_pattern!r}")
-                continue
-
             if exclude_files and (orig_time := exclude_files.get(file)):
                 new_time = file.stat().st_mtime
                 log.info(f"Checking {file.name} ({orig_time=}, {new_time=})")
@@ -168,7 +158,6 @@ class MemFS:
         self,
         limit: int,
         pattern: str,
-        ignores: Sequence[str] = (),
         exclude_files: dict[Path, float] | None = None,
         preload_dict: bool = False,
     ) -> list[FileAttachment]:
@@ -178,7 +167,6 @@ class MemFS:
         Args:
             limit: The maximum number of files to parse.
             pattern: The glob pattern to match files against.
-            ignores: A sequence of fnmatch patterns to ignore.
             exclude_files: A dict of Paths and last modified times.
                 Files will be excluded if their last modified time
                 is equal to the provided value.
@@ -187,7 +175,7 @@ class MemFS:
             List of FileAttachments sorted lexically by path name.
         """
         res = sorted(
-            self.files(limit=limit, pattern=pattern, ignores=ignores, exclude_files=exclude_files),
+            self.files(limit=limit, pattern=pattern, exclude_files=exclude_files),
             key=lambda f: f.path,
         )
         if preload_dict:
--- a/snekbox/memfs.py
+++ b/snekbox/memfs.py
@@ -22,7 +22,13 @@ __all__ = ("MemFS",)
 class MemFS:
     """An in-memory temporary file system."""
 
-    def __init__(self, instance_size: int, root_dir: str | Path = "/memfs") -> None:
+    def __init__(
+        self,
+        instance_size: int,
+        root_dir: str | Path = "/memfs",
+        home: str = "home",
+        output: str = "output",
+    ) -> None:
         """
         Initialize an in-memory temporary file system.
 
@@ -33,10 +39,14 @@ class MemFS:
         Args:
             instance_size: Size limit of each tmpfs instance in bytes.
             root_dir: Root directory to mount instances in.
+            home: Name of the home directory.
+            output: Name of the output directory within home. If empty, uses home.
         """
         self.instance_size = instance_size
         self.root_dir = Path(root_dir)
         self.root_dir.mkdir(exist_ok=True, parents=True)
+        self._home_name = home
+        self._output_name = output
 
         for _ in range(10):
             name = str(uuid4())
@@ -82,12 +92,12 @@ class MemFS:
     @property
     def home(self) -> Path:
         """Path to home directory."""
-        return self.path / "home"
+        return self.path / self._home_name
 
     @property
     def output(self) -> Path:
         """Path to output directory."""
-        return self.home / "output"
+        return self.home / self._output_name
 
     def __enter__(self) -> MemFS:
         return self
@@ -112,7 +122,7 @@ class MemFS:
 
     def files(self, limit: int, pattern: str = "**/*") -> Generator[FileAttachment, None, None]:
         """
-        Yields FileAttachments for files in the MemFS.
+        Yields FileAttachments for files found in the output directory.
 
         Args:
             limit: The maximum number of files to parse.
@@ -134,7 +144,7 @@ class MemFS:
         preload_dict: bool = False,
     ) -> list[FileAttachment]:
         """
-        Return a sorted list of output files found in the MemFS.
+        Return a sorted list of file paths within the output directory.
 
         Args:
             limit: The maximum number of files to parse.
--- a/snekbox/memfs.py
+++ b/snekbox/memfs.py
@@ -137,7 +137,7 @@ class MemFS:
                 yield FileAttachment.from_path(file, relative_to=self.output)
 
     def cleanup(self) -> None:
-        """Unmounts tmpfs."""
+        """Unmount the tmpfs."""
         if self._path is None:
             return
         unmount(self.path)
--- a/snekbox/memfs.py
+++ b/snekbox/memfs.py
@@ -134,7 +134,7 @@ class MemFS:
         preload_dict: bool = False,
     ) -> list[FileAttachment]:
         """
-        Returns a sorted list of output files found in the MemFS.
+        Return a sorted list of output files found in the MemFS.
 
         Args:
             limit: The maximum number of files to parse.
--- a/snekbox/memfs.py
+++ b/snekbox/memfs.py
@@ -62,7 +62,12 @@ class MemFS:
     @property
     def home(self) -> Path:
         """Path to home directory."""
-        return Path(self.path, "home")
+        return self.path / "home"
+
+    @property
+    def output(self) -> Path:
+        """Path to output directory."""
+        return self.home / "output"
 
     def __enter__(self) -> MemFS:
         """Mounts a new tempfs, returns self."""
@@ -79,7 +84,8 @@ class MemFS:
         else:
             raise RuntimeError("Failed to generate a unique tempdir name in 10 attempts")
 
-        self.mkdir("home")
+        self.mkdir(self.home)
+        self.mkdir(self.output)
         return self
 
     def __exit__(
@@ -98,17 +104,17 @@ class MemFS:
         return folder
 
     def attachments(
-        self, max_count: int, pattern: str = "output*"
+        self, max_count: int, pattern: str = "**/*"
     ) -> Generator[FileAttachment, None, None]:
         """Return a list of attachments in the tempdir."""
         count = 0
-        for file in self.home.glob(pattern):
+        for file in self.output.rglob(pattern):
             if count > max_count:
                 log.info(f"Max attachments {max_count} reached, skipping remaining files")
                 break
             if file.is_file():
                 count += 1
-                yield FileAttachment.from_path(file)
+                yield FileAttachment.from_path(file, relative_to=self.output)
 
     def cleanup(self) -> None:
         """Unmounts tmpfs."""
--- a/snekbox/memfs.py
+++ b/snekbox/memfs.py
@@ -81,7 +81,7 @@ class MemFS:
         return self.home / "output"
 
     def __enter__(self) -> MemFS:
-        """Mounts a new tempfs, returns self."""
+        """Mount a new tempfs and return self."""
         for _ in range(10):
             name = str(uuid4())
             try:
--- a/snekbox/memfs.py
+++ b/snekbox/memfs.py
@@ -20,7 +20,7 @@ __all__ = ("MemFS",)
 
 
 class MemFS:
-    """A memory temporary file system."""
+    """An in-memory temporary file system."""
 
     def __init__(self, instance_size: int, root_dir: str | Path = "/memfs") -> None:
         """
--- a/snekbox/memfs.py
+++ b/snekbox/memfs.py
@@ -24,7 +24,7 @@ class MemFS:
 
     def __init__(self, instance_size: int, root_dir: str | Path = "/memfs") -> None:
         """
-        Initialize a memory temporary file system.
+        Initialize an in-memory temporary file system.
 
         Examples:
             >>> with MemFS(1024) as memfs:
--- a/snekbox/memfs.py
+++ b/snekbox/memfs.py
@@ -147,5 +147,6 @@ class MemFS:
         res = sorted(self.files(limit, pattern), key=lambda f: f.path)
         if preload_dict:
             for file in res:
+                # Loads the cached property as attribute
                 _ = file.as_dict
         return res
--- a/snekbox/memfs.py
+++ b/snekbox/memfs.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 import logging
 import warnings
 import weakref
-from collections.abc import Generator
+from collections.abc import Generator, Sequence
 from contextlib import suppress
 from pathlib import Path
 from types import TracebackType
@@ -27,7 +27,7 @@ class MemFS:
         instance_size: int,
         root_dir: str | Path = "/memfs",
         home: str = "home",
-        output: str = "output",
+        output: str = "home",
     ) -> None:
         """
         Initialize an in-memory temporary file system.
@@ -97,7 +97,7 @@ class MemFS:
     @property
     def output(self) -> Path:
         """Path to output directory."""
-        return self.home / self._output_name
+        return self.path / self._output_name
 
     def __enter__(self) -> MemFS:
         return self
@@ -120,21 +120,28 @@ class MemFS:
         folder.chmod(chmod)
         return folder
 
-    def files(self, limit: int, pattern: str = "**/*") -> Generator[FileAttachment, None, None]:
+    def files(
+        self, limit: int, pattern: str = "**/*", exclude_files: Sequence[str] = ()
+    ) -> Generator[FileAttachment, None, None]:
         """
         Yields FileAttachments for files found in the output directory.
 
         Args:
             limit: The maximum number of files to parse.
             pattern: The glob pattern to match files against.
+            exclude_files: Files (relative to home directory) to exclude.
         """
         count = 0
         for file in self.output.rglob(pattern):
+            if str(file.relative_to(self.home)) in exclude_files:
+                log.info(f"Skipping excluded file {file!s}")
+                continue
             if count > limit:
                 log.info(f"Max attachments {limit} reached, skipping remaining files")
                 break
             if file.is_file():
                 count += 1
+                log.info(f"Found file {file!s}")
                 yield FileAttachment.from_path(file, relative_to=self.output)
 
     def files_list(
@@ -142,6 +149,7 @@ class MemFS:
         limit: int,
         pattern: str,
         preload_dict: bool = False,
+        exclude_files: Sequence[str] = (),
     ) -> list[FileAttachment]:
         """
         Return a sorted list of file paths within the output directory.
@@ -150,11 +158,11 @@ class MemFS:
             limit: The maximum number of files to parse.
             pattern: The glob pattern to match files against.
             preload_dict: Whether to preload as_dict property data.
-
+            exclude_files: Files (relative to home directory) to exclude.
         Returns:
             List of FileAttachments sorted lexically by path name.
         """
-        res = sorted(self.files(limit, pattern), key=lambda f: f.path)
+        res = sorted(self.files(limit, pattern, exclude_files), key=lambda f: f.path)
         if preload_dict:
             for file in res:
                 # Loads the cached property as attribute
--- a/snekbox/memfs.py
+++ b/snekbox/memfs.py
@@ -2,8 +2,10 @@
 from __future__ import annotations
 
 import logging
+import warnings
+import weakref
 from collections.abc import Generator
-from functools import cached_property
+from contextlib import suppress
 from pathlib import Path
 from types import TracebackType
 from typing import Type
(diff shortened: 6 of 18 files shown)


snekbox/nsjail.py
```diff
@@ -152,8 +166,28 @@ def python3(
...
 
-        with NamedTemporaryFile() as nsj_log:
-            args = (
+        with NamedTemporaryFile() as nsj_log, MemFS(self.memfs_instance_size) as fs:
+            # Add the temp dir to be mounted as cwd
+            nsjail_args = (
+                # Mount a tmpfs at /dev/shm to support multiprocessing
+                "--mount",
+                # src:dst:fs_type:options
+                f"{fs.shm}:/dev/shm:tmpfs:size={fs.instance_size}",
```
wookie184, line 180: I'm struggling to work out exactly what this does, does this create a new tmpfs (I would guess it does?)? If it does, then surely the `src` does nothing? I think the approach in #113 of defining `/dev/shm` in `snekbox.cfg` and just letting nsjail handle it makes more sense.

snekbox/nsjail.py
```diff
@@ -3,17 +3,21 @@
 import subprocess
 import sys
 import textwrap
-from subprocess import CompletedProcess
 from tempfile import NamedTemporaryFile
 from typing import Iterable
 
 from google.protobuf import text_format
 
+# noinspection PyProtectedMember
```
wookie184, line 11: out of interest, what is pycharm complaining about to make this necessary?

snekbox/nsjail.py
```diff
@@ -163,9 +200,27 @@ def python3(
...
+            match (use_file, c_arg):
+                case (True, _) | (None, False):
+                    args.append("main.py")
+                    # Write the code to a file
+                    with fs.allow_write():
+                        code_path = fs.home / "main.py"
+                        code_path.write_text(code)
+                        log.info(f"Created code file at [{code_path!r}].")
+                case _:
+                    args.append(code)
```
wookie184, line 224: This feels pretty fragile, changing behaviour just based on whether there is a letter `c` in the arguments and having to hardcode support in for timeit isn't great.

Maybe we should make the `input` API parameter optional, and then only add `main.py` if that was passed, otherwise just use the args given (which could contain code). This would be pretty simple behaviour and also flexible. Not sure how we deal with breaking changes for snekbox though.

wookie184 (changes requested):
Not a proper review yet, just some comments from a quick look over.

snekbox/api/resources/eval.py
```diff
@@ -61,7 +59,14 @@ def on_post(self, req: falcon.Request, resp: falcon.Response) -> None:
 
         >>> {
         ...     "stdout": "10000 loops, best of 5: 23.8 usec per loop\n",
-        ...     "returncode": 0
+        ...     "returncode": 0,
+        ...     "attachments": [
+        ...         {
+        ...             "name": "output.png",
+        ...             "mime": "image/png",
```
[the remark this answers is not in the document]
wookie184: Same for size and compression type. If we're only using one compression type then there's no need to include it in the response (and I think supporting multiple compression types is overkill unless there's a real need for it). And with the size, is it really needed?

snekbox/memfs.py
```diff
@@ -0,0 +1,151 @@
...
+        ]
+    )
+    return tmp
+
+
+def unmount_tmpfs(name: str) -> None:
+    """Unmount and remove a tmpfs directory."""
+    tmp = NAMESPACE_DIR / name
+    subprocess.check_call(["umount", str(tmp)])
+    rmtree(tmp, ignore_errors=True)
```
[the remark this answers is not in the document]
ionite34: It seems from my testing it still leaves the folder, should I do a `tmp.unlink()` instead then?

image

snekbox/api/resources/eval.py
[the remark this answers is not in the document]
ionite34: Since we're planning on having a higher file size limit on the snekbox end and doing so on the bot end, it might be a waste of resources to uncompress a very large and well compressed content just to discover it's over our discord-side limit. Some 48MB of repetitive strings could compress to ~4KB, for example.

snekbox/nsjail.py
ionite34: My understanding (from skimming over the barely existant documentation and confusing cpp parser source, is that this mounts a regular folder (in this case backed by tmpfs on the docker side) onto nsjail as a `tmpfs`, so it isn't actually a tmpfs created on the nsjail side, but a fake one? This might be completely wrong, I'm not sure.

However the space here is usable to the full limit of our memfs allocation, and not the memory limit of the instance, so that would suggest it's not using nsjail memory? 

In any case there seems to be 0 difference in whether this is a bindmount (like `home`) or `tmpfs` mount as far as I can tell? I opted for `tmpfs` as it seems more "realistic" as far as the nsjail/python side is concerned on what a `/dev/shm` should be.

snekbox/nsjail.py
ionite34: The fact that `DEBUG` isn't declared in `__all__`, which I guess I understand the intention there?

snekbox/nsjail.py
ionite34: Yeah I agree this isn't ideal, I was just trying to make this work without breaking changes mostly.

The timeit situation mainly stems from the fact that it requires string input
```py
{
	"input": "[i for i in range(100)]",
	"args": ["-m", "timeit"]
}
->
python3 -m timeit "i for i in range(100)"
```
whereas something like `unittest` would requires a module/file
```py
python3 -m unittest main.py
```

So, not sure how we can really decide on snekbox side, maybe this should be something in the input request?

ionite34: Optionally we can just fix `input` to always write a file, and if the user wanted `-c` or `timeit` strings, require that to be in `args` instead?

snekbox/nsjail.py
wookie184: > My understanding (from skimming over the barely existant documentation and confusing cpp parser source), is that this mounts a regular folder (in this case backed by tmpfs on the docker side) onto nsjail as a tmpfs, so it isn't actually a tmpfs created on the nsjail side, but a fake one? This might be completely wrong, I'm not sure.

I did some testing and I think my guess was correct and it just creates a new tmpfs ignoring the source. To demonstrate this you can modify the code to this:
```python
            (fs.shm / "test.txt").write_text("hello world")
            log.info(list(fs.shm.iterdir()))

            try:
                nsjail = subprocess.Popen(
                    args, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, text=True
                )
            except ValueError:
                return EvalResult(args, None, "ValueError: embedded null byte")
            
            log.info(list(fs.shm.iterdir()))
```
And try sending this to snekbox:
```python
import os
open('/dev/shm/thinggggggg.txt', 'w').write('hello world')
print(os.listdir('/dev/shm'))
```
In the logs you'll get something like
```
snekbox_dev  | 2022-11-18 23:23:14,519 |     8 |                 snekbox.nsjail |     INFO | [PosixPath('/memfs/0562f335-93b6-4db6-89d0-1ff580b58921/dev/shm/test.txt')]  
snekbox_dev  | 2022-11-18 23:23:14,520 |     8 |                 snekbox.nsjail |     INFO | [PosixPath('/memfs/0562f335-93b6-4db6-89d0-1ff580b58921/dev/shm/test.txt')] 
```
and in the stdout response from snekbox you'll get
```
['thinggggggg.txt']
```
I also tried replacing `{fs.shm}` with random characters in `f"{fs.shm}:/dev/shm:tmpfs:size={fs.instance_size}"` and it didn't seem to affect how it worked.

> In any case there seems to be 0 difference in whether this is a bindmount (like home) or tmpfs mount as far as I can tell? I opted for tmpfs as it seems more "realistic" as far as the nsjail/python side is concerned on what a /dev/shm should be.

There is a difference (bindmount uses the existing tmpfs), and whilst it doesn't really matter in this case, this current code is confusing (and the code to create `fs.shm` is redundant) so I do think the approach in #113 is better.

ionite34: Okay cool thanks for the testing and info. Yeah I agree in that case the config approach in #113 is probably better then and will be remove the shm mounting in this PR.

snekbox/memfs.py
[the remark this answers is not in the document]
wookie184: Ah, the directory left isn't the tmpfs, it's just the mountpoint directory created in `mount_tmpfs`. I think it would be better to use `tmp.rmdir()` and add a comment to make that clear.

snekbox/memfs.py
```diff
@@ -0,0 +1,154 @@
...
+NAMESPACE_DIR = Path("/memfs")
+NAMESPACE_DIR.mkdir(exist_ok=True)
+NAMESPACE_DIR.chmod(0o711)  # Execute only access for other users
+
+
+def mount_tmpfs(name: str, size: int | str) -> Path:
+    """Create and mount a tmpfs directory."""
+    tmp = NAMESPACE_DIR / name
+    tmp.mkdir()
+    tmp.chmod(0o711)
```
wookie184, line 31: Why is a chmod needed here? (also all the other chmods)

snekbox/memfs.py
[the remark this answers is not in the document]
ionite34: Resolved in [09bbc4f](https://github.com/python-discord/snekbox/pull/159/commits/09bbc4f11f32d91dfa20aeb3da77a5e2a054f49d)

snekbox/memfs.py
ionite34: I think this was a remanent when there used to be exposed access to this folder. Removed in a0c908a. The chmods for the home directory would have to stay because that needs to be read/write.

snekbox/nsjail.py
ionite34: Removed in [9327283](https://github.com/python-discord/snekbox/pull/159/commits/932728308f6773bb525028281ccf1a598d20786b)

snekbox/nsjail.py
wookie184: I think we should probably just add `DEBUG` to `__all__`

ionite34: Added in [893dbd7](https://github.com/python-discord/snekbox/pull/159/commits/893dbd72d6ba2d5d6dfe979795a94d63c3791b97)

snekbox/nsjail.py
ionite34: Implemented new api format in [ba6d0a8](https://github.com/python-discord/snekbox/pull/159/commits/ba6d0a8a10af687393134fc1e9662100ce67df52) which avoids parsing / guessing on snekbox's side in favor of the client providing all arguments and files. 

```json
"args": ["main.py"],
"files": [
    {
        "name": "main.py",
        "content": "print(1)"
    }
]
```

snekbox/api/resources/eval.py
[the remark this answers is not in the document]
ionite34: Removed mime and compression in [ba6d0a8](https://github.com/python-discord/snekbox/pull/159/commits/ba6d0a8a10af687393134fc1e9662100ce67df52)

snekbox/memfs.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def cleanup(self) -> None:
+        """Unmounts tmpfs, releases name."""
+        if self.path is None:
+            return
+        # Remove the path folder
+        unmount_tmpfs(self.name)
+
+        if not self.path.exists():
+            with self.assignment_lock:
```
MarkKoz, line 128: Why is a lock necessary?

snekbox/nsjail.py
```diff
@@ -163,37 +193,49 @@ def python3(
                 "--",
                 self.config.exec_bin.path,
                 *self.config.exec_bin.arg,
-                *py_args,
-                code,
-            )
+                # Filter out empty strings (causes issues with python cli)
```
MarkKoz, line 196: Only empty strings at the start should be removed. Empty strings elsewhere could conceivably be valid arguments for the script.

snekbox/nsjail.py
```diff
@@ -163,37 +193,49 @@ def python3(
...
+                *(arg for arg in py_args if arg),
+            ]
+
+            # Write files if any
+            for file in files:
+                file.save_to(fs.home)
+                log.info(f"Created file at [{(fs.home / file.name)!r}].")
```
MarkKoz, line 203: The brackets seem unnecessary; the path is already going to be quoted.
```suggestion
                log.info(f"Created file at {(fs.home / file.name)!r}.")
```

snekbox/nsjail.py
```diff
@@ -163,37 +193,49 @@ def python3(
...
 
             msg = "Executing code..."
             if DEBUG:
-                msg = f"{msg[:-3]}:\n{textwrap.indent(code, '    ')}\nWith the arguments {args}."
+                msg = f"{msg[:-3]}: With the arguments {args}."
```
MarkKoz, line 207:
```suggestion
                msg = f"{msg[:-3]} with the arguments {args}."
```

snekbox/nsjail.py
```diff
@@ -129,16 +137,24 @@ def _consume_stdout(self, nsjail: subprocess.Popen) -> str:
...
         The `nsjail_args` passed will be used to override the values in the NsJail config.
         These arguments are only options for NsJail; they do not affect Python's arguments.
 
         `py_args` are arguments to pass to the Python subprocess before the code,
-        which is the last argument. By default, it's "-c", which executes the code given.
+        which is the last argument.
+
+        If `use_file` is True, `code` will be written to a file the last argument will be
+        the file name `main.py`. If False, `code` will be passed as the last argument.
+        The default of None will use a file unless `c` is in `py_args`.
```
MarkKoz, line 157: This seems out of date.

snekbox/snekio.py
```diff
@@ -0,0 +1,108 @@
...
+        if any(set(part) == {"."} for part in parts):
+            raise IllegalPathError(f"File path '{name}' may not use traversal ('..')")
+
+        match data.get("content-encoding"):
+            case "base64":
+                content = b64decode(data["content"])
+            case None | "utf-8" | "":
+                content = data["content"].encode("utf-8")
+            case _:
+                raise ParsingError(f"Unknown content encoding '{data['content-encoding']}'")
```
MarkKoz, line 62: Let's not overcomplicate this. Similar to how the response always uses b64 for files, the request should also always use b64 for files.

snekbox/memfs.py
```diff
@@ -0,0 +1,137 @@
...
+        folder.chmod(chmod)
+        return folder
+
+    def attachments(
+        self, max_count: int, max_size: int | None = None
+    ) -> Generator[FileAttachment, None, None]:
+        """Return a list of attachments in the tempdir."""
+        count = 0
+        # Look for any file starting with `output`
+        for file in self.home.glob("output*"):
```
MarkKoz, line 112: Thoughts on requiring writing to an "output" directory rather than requiring file names to be prefixed with `output`?

snekbox/nsjail.py
```diff
@@ -129,16 +137,24 @@ def _consume_stdout(self, nsjail: subprocess.Popen) -> str:
         return "".join(output)
 
     def python3(
-        self, code: str, *, nsjail_args: Iterable[str] = (), py_args: Iterable[str] = ("-c",)
-    ) -> CompletedProcess:
+        self,
+        py_args: Iterable[str],
+        files: Iterable[FileAttachment] = (),
+        *,
```
MarkKoz, line 143: The placement of this `*` seems arbitrary. It would make sense for it to be before `py_args`, but I would be fine if it was just removed entirely at this point.

snekbox/nsjail.py
```diff
@@ -152,8 +168,22 @@ def python3(
...
+            nsjail_args = (
+                # Mount `home` with Read/Write access
+                "--bindmount",
+                f"{fs.home}:home",
+                # Set cwd to temp dir
+                "--cwd",
+                "home",
+                # Some packages rely on the HOME env variable
+                "--env",
+                "HOME=home",
```
MarkKoz, line 182: Why not set these in the config file instead?

snekbox/__main__.py
```diff
@@ -37,7 +37,7 @@ def parse_args() -> argparse.Namespace:
 def main() -> None:
     """Evaluate Python code through NsJail."""
     args = parse_args()
-    result = NsJail().python3(args.code, nsjail_args=args.nsjail_args, py_args=args.py_args)
+    result = NsJail().python3(py_args=[*args.py_args, args.code], nsjail_args=args.nsjail_args)
```
MarkKoz, line 40: `py_args` needs to default to `-c` here to preserve the original behaviour, since `python3()` no longer has it as a default. (I think this is the most convenient default when using it via the CLI for testing).

MarkKoz (changes requested):

snekbox/snekio.py
```diff
@@ -0,0 +1,108 @@
...
+
+    @classmethod
+    def from_dict(cls, data: dict[str, str]) -> FileAttachment:
+        """Convert a dict to an attachment."""
+        name = data["name"]
+        path = Path(name)
+        parts = path.parts
+
+        if path.is_absolute() or set(parts[0]) & {"\\", "/"}:
+            raise IllegalPathError(f"File path '{name}' must be relative")
```
MarkKoz, line 51: `is_absolute` covers the `/` case, and `"\\"` is of no concern since it's not considered absolute on Posix.

snekbox/snekio.py
```diff
@@ -0,0 +1,108 @@
...
+
+        if any(set(part) == {"."} for part in parts):
+            raise IllegalPathError(f"File path '{name}' may not use traversal ('..')")
```
MarkKoz, line 54: This only checks a single dot. Shouldn't it also check `..`? I wonder if there is a more convenient way to detect this. Maybe https://stackoverflow.com/a/45190125 ?

MarkKoz (changes requested):

snekbox/memfs.py
ionite34: I was thinking there might be issues if we had multiple threads? Though I suppose we don't have that currently. I don't think it hurts to make it thread safe for the time being though?

tests/test_nsjail.py
```diff
@@ -272,14 +336,14 @@ def test_large_output_is_truncated(self):
         self.assertEqual(output, chunk * expected_chunks)
 
     def test_nsjail_args(self):
-        args = ("foo", "bar")
+        args = ["foo", "bar"]
```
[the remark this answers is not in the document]
ionite34: It's due to the fact I've changed the call within the function to a list and not a tuple since I had to do a couple extend/appends. Wasn't sure whether to cast back to a tuple or just leave it and ended up just leaving it. Do you think it's better if we cast it to a tuple at the end?

docker-compose.yml
```diff
@@ -8,7 +8,7 @@ services:
     image: ghcr.io/python-discord/snekbox${IMAGE_SUFFIX:--venv:dev}
     pull_policy: never
     ports:
-     - 8060:8060
+     - "8060:8060"
```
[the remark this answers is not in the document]
ionite34: Docker recommends to only use strings for port numbers https://docs.docker.com/compose/compose-file/compose-file-v3/#short-syntax-1

snekbox/api/resources/eval.py
```diff
@@ -38,8 +49,6 @@ def on_post(self, req: falcon.Request, resp: falcon.Response) -> None:
         Evaluate Python code and return stdout, stderr, and the return code.
 
         A list of arguments for the Python subprocess can be specified as `args`.
-        Otherwise, the default argument "-c" is used to execute the input code.
-        The input code is always passed as the last argument to Python.
```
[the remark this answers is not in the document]
ionite34: Seems reasonable yeah, I'll look into compatible support for "input"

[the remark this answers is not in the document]
ionite34: should we have args implicitly include an "-c" default if "input" is non-empty?

tests/test_nsjail.py
```diff
@@ -99,7 +103,7 @@ def f():
         self.assertEqual(result.stderr, None)
 
     def test_read_only_file_system(self):
-        for path in ("/", "/etc", "/lib", "/lib64", "/snekbox", "/usr"):
+        for path in ("/etc", "/lib", "/lib64", "/snekbox", "/usr"):
```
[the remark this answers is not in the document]
ionite34: Seems accidental. Re-added in [291119d](https://github.com/python-discord/snekbox/pull/159/commits/291119dc45817a01f29afcde0328a53f9102df3c)

tests/test_nsjail.py
```diff
@@ -113,6 +117,64 @@ def test_read_only_file_system(self):
...
+                    f.write(b'1' * 2048)
+            """
+        ).strip()
+
+        result = self.nsjail.python3(code)
+        self.assertEqual(result.returncode, 1)
+        self.assertIn("No space left on device", result.stdout)
+        self.assertEqual(result.stderr, None)
+        # Restore the original size
+        self.nsjail.memfs_instance_size = backup
```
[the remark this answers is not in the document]
ionite34: Make instance size lower global for the tests in [a1c614d](https://github.com/python-discord/snekbox/pull/159/commits/a1c614d7e950a6329bfaeec7ef1467553bd96a37)

snekbox/snekio.py
```diff
@@ -0,0 +1,89 @@
...
+
+    @classmethod
+    def from_path(cls, file: Path, max_size: int | None = None) -> FileAttachment:
+        """Create an attachment from a path."""
+        size = file.stat().st_size
+        if max_size is not None and size > max_size:
+            raise AttachmentError(
+                f"File {file.name} too large: {sizeof_fmt(size)} "
+                f"exceeds the limit of {sizeof_fmt(max_size)}"
+            )
```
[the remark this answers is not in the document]
ionite34: Removed file size limit and associated errors, added configurable file parse time limit in [73a5501](https://github.com/python-discord/snekbox/pull/159/commits/73a5501e04706e190ee7d97ac006924b0e90463a)

snekbox/snekio.py
ionite34: Made request accept base64 only, removed `content-encoding` parameter, in [4787a94](https://github.com/python-discord/snekbox/pull/159/commits/4787a9497be4f460e028031ddad4cd9d8388b84f)

snekbox/snekio.py
```diff
@@ -0,0 +1,61 @@
...
+        return mimetypes.guess_type(self.name)[0]
+
+    @property
+    def size(self) -> int:
+        """Size of the attachment."""
+        return len(self.content)
+
+    def to_dict(self) -> dict[str, str]:
+        """Convert the attachment to a dict."""
+        cmp = zlib.compress(self.content)
```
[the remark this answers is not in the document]
ionite34: I've removed the zlib compression completely and content-encoding parsing in [4787a94](https://github.com/python-discord/snekbox/pull/159/commits/4787a9497be4f460e028031ddad4cd9d8388b84f)

I think since compression doesn't seem essential at this point, for the file sizes we're dealing with, (and won't change API if we implement server-layer compression), I'll probably leave that for a future PR.

config/snekbox.cfg
```diff
@@ -104,11 +104,13 @@ mount {
     rw: false
 }
 
+rlimit_fsize: 134217728
```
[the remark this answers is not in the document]
ionite34: Set to -1 in [80d4c45](https://github.com/python-discord/snekbox/pull/159/commits/80d4c45ae6bd152dd7c1dfcfbaa1995dd0018954)

snekbox/api/resources/eval.py
```diff
@@ -53,33 +62,57 @@ def on_post(self, req: falcon.Request, resp: falcon.Response) -> None:
...
         Response format:
 
         >>> {
-        ...     "stdout": "10000 loops, best of 5: 23.8 usec per loop\n",
-        ...     "returncode": 0
+        ...     "stdout": "10000 loops, best of 5: 23.8 usec per loop",
+        ...     "returncode": 0,
+        ...     "attachments": [
+        ...         {
+        ...             "name": "output.png",
```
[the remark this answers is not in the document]
ionite34: Renamed to path in 4787a94

snekbox/api/resources/eval.py
```diff
@@ -53,33 +62,57 @@ def on_post(self, req: falcon.Request, resp: falcon.Response) -> None:
...
-        ...     "input": "[i for i in range(1000)]",
-        ...     "args": ["-m", "timeit"] # This is optional
+        ...    "args": ["-c", "print('Hello')"]
+        ... }
+
+        >>> {
+        ...    "args": ["main.py"],
+        ...    "files": [
+        ...        {
+        ...            "name": "main.py",
```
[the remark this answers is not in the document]
ionite34: Renamed to path in 4787a94

snekbox/api/resources/eval.py
```diff
@@ -53,33 +62,57 @@ def on_post(self, req: falcon.Request, resp: falcon.Response) -> None:
...
+        ...     "stdout": "10000 loops, best of 5: 23.8 usec per loop",
+        ...     "returncode": 0,
+        ...     "attachments": [
+        ...         {
+        ...             "name": "output.png",
+        ...             "mime": "image/png",
+        ...             "size": 57344,
+        ...             "compression": "zlib",
+        ...             "content": "eJzzSM3...="  # Base64-encoded
+        ...         }
```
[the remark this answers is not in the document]
ionite34: Updated example along with changes in 4787a94

tests/test_nsjail.py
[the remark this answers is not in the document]
MarkKoz: Thanks for clarifying. No objections to keeping it a list.

docker-compose.yml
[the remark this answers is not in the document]
MarkKoz: Thanks. YAML is very cool and good.

config/snekbox.cfg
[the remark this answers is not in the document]
MarkKoz: Ah, I did not realise NsJail had a default. Unlimited is fine, I think... Ultimately we're concerned without overall disk usage being limited. If that's somehow bypassed or misconfigured, I don't see how a malicious user creating 1 large file (when rlimit is -1) vs. many small files (when rlimit is set) would have a significant impact.

tests/test_nsjail.py
```diff
@@ -63,9 +85,12 @@ def test_subprocess_resource_unavailable(self):
             """
         ).strip()
 
-        result = self.nsjail.python3(code)
+        result = self.eval_file(code)
         self.assertEqual(result.returncode, 1)
         self.assertIn("Resource temporarily unavailable", result.stdout)
+        # Also expect n-1 processes to be opened
+        res = result.stdout.split("T")[0].strip().split("\n")  # ['2', '3', '4'] for n=4
+        self.assertEqual(res, [*map(str, range(2, max_pids + 1))])
```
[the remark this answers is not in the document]
MarkKoz: Oh, this is checking the PIDs being printed! I didn't see that those get printed at first; I got lost trying to understand the string parsing happening here. A comment explaining that PIDs are being extracted from stdout would help.

snekbox/api/resources/eval.py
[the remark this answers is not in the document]
MarkKoz: If `input` exists, it should become the last Python argument. If `input` exists and `args` does not exist, then `args` should default to `["-c"]` (in addition to the `input` becoming the last Python argument).

snekbox/memfs.py
MarkKoz: Gunicorn is configured with multiple workers, though right now I cannot recall if those are threads or processes. My vote is for simplicity, so a lock shouldn't be included if it's not necessary. Can you investigate further? If it is needed, then there should also be a test for whatever bad situation this lock prevents.

snekbox/nsjail.py
```diff
@@ -205,8 +205,9 @@ def python3(
                 "--",
                 self.config.exec_bin.path,
                 *self.config.exec_bin.arg,
-                # Filter out empty strings (causes issues with python cli)
-                *(arg for arg in py_args if arg),
+                # Filter out empty strings at start of iterable
+                # (causes issues with python cli)
+                *(arg for i, arg in enumerate(py_args) if (arg or i > 0)),
```
MarkKoz, line 209: This doesn't cover the case of multiple empty strings at the start. I also think writing a test for this would be beneficial.

MarkKoz (changes requested):

snekbox/nsjail.py
```diff
@@ -227,13 +239,14 @@ def python3(
...
+                    parse_files,
+                    (fs, self.files_limit, self.files_pattern),
+                    timeout=self.files_timeout,
                 )
                 log.info(f"Found {len(attachments)} attachments.")
-            except AttachmentError as err:
-                log.info(f"Failed to parse attachments: {err}")
+            except TimeoutError as err:
+                log.info(f"Exceeded time limit in parsing attachments: {err}")
                 return EvalResult(args, returncode, f"AttachmentError: {err}")
```
MarkKoz, line 250: The `AttachmentError: ` prefix seems misleading because no such exception type still exists. I don't think it's necessary.

snekbox/nsjail.py
```diff
@@ -227,13 +239,14 @@ def python3(
...
+                attachments = timed(
+                    parse_files,
+                    (fs, self.files_limit, self.files_pattern),
+                    timeout=self.files_timeout,
                 )
                 log.info(f"Found {len(attachments)} attachments.")
-            except AttachmentError as err:
-                log.info(f"Failed to parse attachments: {err}")
+            except TimeoutError as err:
+                log.info(f"Exceeded time limit in parsing attachments: {err}")
```
MarkKoz, line 249: I have mixed feelings towards having a timeout. I'm leaning towards keeping it simple and not having a timeout. It should be possible to prevent long parsing times by just limiting the amount of files, for which we already have a setting.

Furthermore, it seems unintuitive that the process may have an OK exit code with some stdout available, but it's getting overwritten by an error, and the HTTP response is still 200. This is also a concern I had with the previous code that was erroring out when a file size limit was exceeded.

What are everyone's thoughts on having a timeout for file parsing? Is it beneficial given the other measures we have in place?

snekbox/nsjail.py
```diff
@@ -38,16 +48,18 @@ def __init__(
         max_output_size: int = 1_000_000,
         read_chunk_size: int = 10_000,
         memfs_instance_size: int = 48 * 1024 * 1024,
-        max_attachments: int = 100,
-        max_attachment_size: int | None = None,
+        files_limit: int = 100,
+        files_timeout: float = 15,
+        files_pattern: str = "output*",
```
MarkKoz, line 53: I think it's a good idea that this is configurable now.

MarkKoz (changes requested):

snekbox/nsjail.py
ionite34: My main concern is, if we allow nested directories in outputs, someone could create a very deeply nested directory as a denial of service attack. 

In general this is a potentially lengthy part of server side code execution that can be entirely controlled by the user.

ionite34: I agree that the response feels unintuitive though. Maybe it can just be a 500 error? This is more of a catastrophic fail-safe than something that is going to come up often.

snekbox/nsjail.py
ionite34: Moved cwd and home env to config file in [c30c0eb](https://github.com/python-discord/snekbox/pull/159/commits/c30c0eb443f3ab759e9dfcbfee5bac6f4c428048)

snekbox/nsjail.py
ionite34: Removed `*` in [4787a94](https://github.com/python-discord/snekbox/pull/159/commits/4787a9497be4f460e028031ddad4cd9d8388b84f)

snekbox/nsjail.py
ionite34: I have changed the returncode of the TimeoutError to `None` to be in-line with the previous snekbox stdout errors.

snekbox/nsjail.py
ionite34: Changed in [9400107](https://github.com/python-discord/snekbox/pull/159/commits/9400107e00d37b1809dd1ff8276a6b5bbba9dfdd)

snekbox/snekio.py
ionite34: Implemented the pathlib solution since [00b7e3b](https://github.com/python-discord/snekbox/pull/159/commits/00b7e3bd13b7914a44c5af00ff6f8da9b48fa9c5) as detailed in the stackoverflow post. Should be better since now we only disallow traversal beyond root.

snekbox/snekio.py
ionite34: Only using `is_absolute` since [00b7e3b](https://github.com/python-discord/snekbox/pull/159/commits/00b7e3bd13b7914a44c5af00ff6f8da9b48fa9c5)

snekbox/nsjail.py
ionite34: Created function `iter_lstrip` for this task, and additional unit tests since [621c835](https://github.com/python-discord/snekbox/pull/159/commits/621c835f28cd3fcd1e5c3cccae9b5057647f9fa1)

snekbox/api/resources/eval.py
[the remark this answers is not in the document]
ionite34: Honestly after attempting an implementation, this seems to just increase complexity and error-prone code (like the fact that our specification will be an 'either-or-but-at-least-one' requirement of `args` and `input`. The initial inclusion of `input` was already largely redundant as it is directly appended to `args`, so `args` can achieve all of the previous usages of `input` but without the ambiguous parsing and dynamic default of "-c" that we do on our end. 

In the future, if we support requests specifying user input via `sys.argv`, the name occupied by `input` will be more ambiguous and confusing. 

In general the deprecation of `input` at this stage seems like a reasonable path. If we truly need to support users beyond python bot, imo the usual case is maintaining a versioned endpoint that retains previous behavior and specifications fully, instead of retaining it as an optional parameter in the same endpoint (Which I've never really seen as a path taken by any major public API).

For the CLI interface, the simplicity in typing the code directly could be made, but for an API, we're retaining a parameter (input) just to be later conditionally appended to an already existing parameter (args) which has a default that conditionally exists based on whether the new parameter is empty or not. With all of this being able to be avoided from the requester supplying a single explicit "-c" in `args`.

Considering the additional keys in the response is already potentially breaking, and our numerous breaking library-side changes (such as the return types and arguments of `python3`), I don't really see a library side breaking concern more than any other change. We should increment a major version, and downstream users can upgrade when they're ready.

snekbox/__main__.py
ionite34: I think the argparse default here already makes `-c` the default? The previous unit tests for main also work fine.

tests/test_nsjail.py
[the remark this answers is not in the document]
ionite34: Refactored and added comments to make this clearer in [2b20b85](https://github.com/python-discord/snekbox/pull/159/commits/2b20b854b027c1ba27aa85b1dcef96c4bfa7b6b9)

snekbox/memfs.py
ionite34: It seems the `path.mkdir()` in `mount_tmpfs` should already be sufficient as the python unix implementation appears to be free from race conditions.

Removed set and semaphore in [8c897dc](https://github.com/python-discord/snekbox/pull/159/commits/8c897dc0e86f4c6cb5e09e376c6aa40d4093ed7d) in favor of catching the `FileExistsError` instead.

Added tests in [b74b813](https://github.com/python-discord/snekbox/pull/159/commits/b74b813ae6ac2ebf374cc12bb0e9689f49776f2e) to ensure this new approach works fine in multithread conditions.

snekbox/memfs.py
```diff
@@ -0,0 +1,137 @@
...
+    assignment_lock = BoundedSemaphore(1)
+    assigned_names: set[str] = set()  # Pool of tempdir names in use
+
+    def __init__(self, instance_size: int) -> None:
+        """
+        Create a temporary directory using tmpfs.
+
+        size: Size limit of each tmpfs instance in bytes
+        """
+        self.path: Path | None = None
```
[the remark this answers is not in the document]
ionite34: Changed in [8c897dc](https://github.com/python-discord/snekbox/pull/159/commits/8c897dc0e86f4c6cb5e09e376c6aa40d4093ed7d) to raise a `RuntimeError` on most attribute access when not initialized through `__enter__`

snekbox/memfs.py
ionite34: Hm, I think that sounds reasonable. I was originally trying to avoid creating any other directories in the instance /home since now there's a potential FileExistsError if an user provides a file called `output` in the request. Though I suppose we could check for that as a validation?

snekbox/nsjail.py
ionite34: Updated in [f12f8f3](https://github.com/python-discord/snekbox/pull/159/commits/f12f8f3915620cf5f20ef5c36e488517a4e73e4b)

snekbox/memfs.py
```diff
@@ -0,0 +1,137 @@
...
+        [
+            "mount",
+            "-t",
+            "tmpfs",
+            "-o",
+            f"size={size}",
+            "tmpfs",
+            str(tmp),
+        ]
+    )
```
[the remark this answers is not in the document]
ionite34: Refactored to use syscall for mount and unmount in [0387793](https://github.com/python-discord/snekbox/pull/159/commits/0387793222cb1c0feb5644ba8ba8c26caa84c7b2)

snekbox/memfs.py
ionite34: Implemented since [78b4b6a](https://github.com/python-discord/snekbox/pull/159/commits/78b4b6af18a40db3162aad56eb726a26c5a74e8c)

snekbox/snekio.py
```diff
@@ -0,0 +1,89 @@
...
+        content = b64encode(cmp).decode("ascii")
+        return {
+            "name": self.name,
+            "size": self.size,
```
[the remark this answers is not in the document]
ionite34: For our usage this saves python bot having to decode a base64 content string if size already exceeds our limit, which is not a huge optimization but could be a useful information which is free for us to include.

tests/test_libmount.py
```diff
@@ -0,0 +1,135 @@
...
+    def test_unmount_flags(self):
+        """Test unmount flags."""
+        flags = [
+            libmount.UnmountFlags.MNT_FORCE,
+            libmount.UnmountFlags.MNT_DETACH,
+            libmount.UnmountFlags.UMOUNT_NOFOLLOW,
+        ]
+        for flag in flags:
+            with self.subTest(flag=flag), self.get_mount() as path:
+                libmount.unmount(path, flag)
```
MarkKoz, line 75: What is the assertion here?

tests/test_nsjail.py
```diff
@@ -135,25 +223,27 @@ def test_sigsegv_returns_139(self):  # In honour of Juan.
...
+        result = self.eval_file(code)
         self.assertEqual(result.returncode, 139)
         self.assertEqual(result.stdout, "")
         self.assertEqual(result.stderr, None)
 
     def test_null_byte_value_error(self):
-        result = self.nsjail.python3("\0")
+        # This error does not occur without -c, where it
+        # would be a normal SyntaxError.
+        result = self.nsjail.python3(["-c", "\0"])
```
MarkKoz, line 234: This can use the `eval_code` utility, right?

snekbox/memfs.py
```diff
@@ -0,0 +1,148 @@
...
+        self.instance_size = instance_size
+        self._path: Path | None = None
+        self.root_dir: Path = Path(root_dir)
+        self.root_dir.mkdir(exist_ok=True, parents=True)
+
+    @cached_property
+    def path(self) -> Path:
+        """Returns the path of the MemFS."""
+        if self._path is None:
+            raise RuntimeError("MemFS accessed before __enter__.")
```
MarkKoz, line 65: Sorry to sort of "go back" on this, but it dawned on me: why is this being set in `__enter__` instead of `__init__`? Most libraries I've seen support using an object without a context manager (requiring a manual call to some public cleanup function). Wouldn't it make more sense for all of `__enter__` to just be in `__init__` instead? This current situation of being able to instantiate an object in an invalid state just feels really wrong. On the other hand, having side effects when instantiation an object also doesn't sound great, but the former seems like the worse of the two.

I don't want to block the PR on this. This is all internal stuff we can revisit at a later time if you prefer.

snekbox/nsjail.py
```diff
@@ -129,16 +165,18 @@ def _consume_stdout(self, nsjail: subprocess.Popen) -> str:
...
 
-        The `nsjail_args` passed will be used to override the values in the NsJail config.
-        These arguments are only options for NsJail; they do not affect Python's arguments.
-
-        `py_args` are arguments to pass to the Python subprocess before the code,
-        which is the last argument. By default, it's "-c", which executes the code given.
+        Args:
+            py_args: Arguments passed to python3.
+            files: FileAttachments to be written to the sandbox prior to run.
+            nsjail_args: Overrides options in the NsJail config, added to NsJail call.
```
MarkKoz, line 179:
```suggestion
            py_args: Arguments to pass to Python.
            files: FileAttachments to write to the sandbox prior to running Python.
            nsjail_args: Overrides for the NsJail configuration.
```

snekbox/snekio.py
```diff
@@ -0,0 +1,101 @@
...
+        return len(self.content)
+
+    def as_bytes(self) -> bytes:
+        """Return the attachment as bytes."""
+        if isinstance(self.content, bytes):
+            return self.content
+        return self.content.encode("utf-8")
+
+    def save_to(self, directory: Path | str) -> None:
+        """Save the attachment to a path directory."""
```
MarkKoz, line 84:
```suggestion
        """Write the attachment to a file in `directory`."""
```

MarkKoz: The `directory` parameter should be included in the docstring.

snekbox/nsjail.py
```diff
@@ -2,26 +2,41 @@
...
+_T = TypeVar("_T")
+
 # [level][timestamp][PID]? function_signature:line_no? message
 LOG_PATTERN = re.compile(
     r"\[(?P<level>(I)|[DWEF])\]\[.+?\](?(2)|(?P<func>\[\d+\] .+?:\d+ )) ?(?P<msg>.+)"
 )
 
 
+def iter_lstrip(iterable: Iterable[_T]) -> Generator[_T, None, None]:
+    """Removes leading falsy objects from an iterable."""
```
MarkKoz, line 31:
```suggestion
    """Remove leading falsy objects from an iterable."""
```

MarkKoz (changes requested):
I'm concerned that I'm leaving too many comments on non-blockers. Let me know if you'd prefer to defer some comments to a future PR (which doesn't necessarily have to be implemented by you). I don't want to create a draining review cycle for you.

snekbox/nsjail.py
```diff
@@ -19,18 +20,35 @@
...
     files_limit: int,
     files_pattern: str,
 ) -> list[FileAttachment]:
-    """Parse files in a MemFS."""
+    """
+    Parse files in a MemFS.
+
+    Returns:
+        List of FileAttachments sorted lexically by path name.
+    """
```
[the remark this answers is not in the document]
MarkKoz: Okay this comment is a bit outdated now that I noticed you changed the docstring style. Still, I think this could just be changed to the one liner I suggested above.

tests/test_nsjail.py
```diff
@@ -361,11 +361,18 @@ def test_nsjail_args(self):
...
-        result = self.nsjail.python3(args)
-
-        self.assertEqual(result.returncode, 0)
-        self.assertEqual(result.args[-2:], args)
+        expected = ["-m", "timeit"]
+        args = [
+            ["", "-m", "timeit"],
+            ["", "", "-m", "timeit"],
+            ["", "", "", "-m", "timeit"],
+        ]
```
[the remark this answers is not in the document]
ionite34: Included since [2da0ddf](https://github.com/python-discord/snekbox/pull/159/commits/2da0ddf14625cfd8037c198881ab789827746c40)

tests/test_nsjail.py
ionite34: Yup, using `eval_code` with [fca97ce](https://github.com/python-discord/snekbox/pull/159/commits/fca97cecea5040dac8ab02046cffac0713f78dd2)

tests/test_libmount.py
```diff
@@ -0,0 +1,135 @@
...
+from unittest import TestCase
+from uuid import uuid4
+
+from snekbox import libmount
+
+
+class LibMountTests(TestCase):
+    def setUp(self):
+        self.temp_dir = Path("/tmp/snekbox-test")
+        self.temp_dir.mkdir(exist_ok=True, parents=True)
```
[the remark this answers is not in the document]
ionite34: Changed to use `setUpClass` and `tearDownClass` and cleanups in [f4494c3](https://github.com/python-discord/snekbox/pull/159/commits/f4494c307eb5cf4200fd678dff90f1c2b5a5775e)

snekbox/snekio.py
```diff
@@ -0,0 +1,101 @@
...
+class ParsingError(AttachmentError):
+    """Raised when an incoming file cannot be parsed."""
+
+
+class IllegalPathError(ParsingError):
+    """Raised when an attachment has an illegal path."""
+
+
+@dataclass
+class FileAttachment(Generic[T]):
```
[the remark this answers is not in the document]
ionite34: Refactored as non-generic in [407e6b2](https://github.com/python-discord/snekbox/pull/159/commits/407e6b2079ea9f73f52e4972147620d765d96349)

snekbox/snekio.py
```diff
@@ -0,0 +1,101 @@
...
+    @property
+    def size(self) -> int:
+        """Size of the attachment."""
+        return len(self.content)
+
+    def as_bytes(self) -> bytes:
+        """Return the attachment as bytes."""
+        if isinstance(self.content, bytes):
+            return self.content
+        return self.content.encode("utf-8")
```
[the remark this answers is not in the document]
ionite34: Refactored as non-generic in [407e6b2](https://github.com/python-discord/snekbox/pull/159/commits/407e6b2079ea9f73f52e4972147620d765d96349)

snekbox/snekio.py
```diff
@@ -0,0 +1,101 @@
...
+        file = Path(directory, self.path)
+        # Create directories if they don't exist
+        file.parent.mkdir(parents=True, exist_ok=True)
+        if isinstance(self.content, str):
+            file.write_text(self.content, encoding="utf-8")
+        else:
+            file.write_bytes(self.content)
+
+    @cached_property
+    def json(self) -> dict[str, str]:
```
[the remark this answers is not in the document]
ionite34: I've renamed to `as_dict` since I feel the fact that this is a property and not a method, it might be confusing to share the same name `to_dict` as the other method ones. [fabaae9](https://github.com/python-discord/snekbox/pull/159/commits/fabaae9472321ec35a0736b96ea4af9b72d30eae)

snekbox/snekio.py
```diff
@@ -0,0 +1,101 @@
...
+class FileAttachment(Generic[T]):
+    """A file attachment."""
+
+    path: str
+    content: T
+
+    @classmethod
+    def from_dict(cls, data: dict[str, str]) -> FileAttachment[bytes]:
+        """Convert a dict to an attachment."""
+        path = safe_path(data["path"])
```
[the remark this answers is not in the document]
ionite34: I think the only invalid path is a single `/` on linux? And that would be caught by as an absolute path. File writing is also in a `try` block with an associated `OSError` report.

Though I have added handling for absolute paths (any path beginning with `/` or `\0`) and unit tests for these in [7e7c0d6](https://github.com/python-discord/snekbox/pull/159/commits/7e7c0d6f5465f965b3ee31fc4f723ec23cc5ad94)

I am not sure if we can do traversal checks with regex. And I think we should still keep the current pathlib checks for absolute and traversal (since the json schema validator is not used for `__main__` calls and library use)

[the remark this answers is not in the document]
ionite34: Added additional unit tests in [dffd9c8](https://github.com/python-discord/snekbox/pull/159/commits/dffd9c8100230d6e8df8c0261fbd5ae8dbd0ceb2)

snekbox/libmount.py
```diff
@@ -0,0 +1,81 @@
...
+
+    result: int = libc.mount(
+        str(source).encode(), str(target).encode(), fs.encode(), 0, kwargs.encode()
+    )
+    if result < 0:
+        errno = ctypes.get_errno()
+        raise OSError(errno, f"Error mounting {target}: {os.strerror(errno)}")
+
+
+def unmount(target: Path | str, flags: UnmountFlags | int = UnmountFlags.MNT_DETACH) -> None:
```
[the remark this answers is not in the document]
ionite34: I think that's better yeah, renamed to `filesystem` in [e72032f](https://github.com/python-discord/snekbox/pull/159/commits/e72032fc3090d6e5d5d961a3aafaeec4cce787d7)

snekbox/snekio.py
```diff
@@ -0,0 +1,101 @@
...
+        content = b64decode(data.get("content", ""))
```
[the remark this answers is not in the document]
ionite34: Added error handling in [0442a7c](https://github.com/python-discord/snekbox/pull/159/commits/0442a7cec24bff9985fb8cf59b0fb2604544eaa8), additional base64 unit tests in [dffd9c8](https://github.com/python-discord/snekbox/pull/159/commits/dffd9c8100230d6e8df8c0261fbd5ae8dbd0ceb2)

snekbox/api/resources/eval.py
```diff
@@ -53,33 +65,57 @@ def on_post(self, req: falcon.Request, resp: falcon.Response) -> None:
...
         try:
-            result = self.nsjail.python3(code, py_args=args)
+            result = self.nsjail.python3(
+                py_args=req.media["args"],
+                files=[FileAttachment.from_dict(file) for file in req.media.get("files", [])],
+            )
+        except IllegalPathError as e:
+            raise falcon.HTTPBadRequest(
+                title="Request file path failed validation", description=str(e)
+            )
```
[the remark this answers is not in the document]
ionite34: I've added schema level validation for `\0` and paths starting with `/`, though I think we'll still need this for the traversal checks and for `__main__` usage.

snekbox/nsjail.py
```diff
@@ -35,11 +50,32 @@ def __init__(
...
+        Initialize NsJail.
+
+        Args:
+            nsjail_path: Path to the NsJail binary.
+            config_path: Path to the NsJail configuration file.
+            max_output_size: Maximum size of the output in bytes.
+            read_chunk_size: Size of the read buffer in bytes.
+            memfs_instance_size: Size of the tmpfs instance in bytes.
+            files_limit: Maximum number of files to parse for attach.
+            files_timeout: Maximum time in seconds to wait for files to be written / read.
```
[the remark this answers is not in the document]
ionite34: Should be yeah, clarified in [2ba9431](https://github.com/python-discord/snekbox/pull/159/commits/2ba94312e60300b136b55781c8f8e3149e8bb0ad)

snekbox/nsjail.py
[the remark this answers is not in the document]
ionite34: Refactored `parse_files` as an instance function `files_list` of `MemFS` in [8272f88](https://github.com/python-discord/snekbox/pull/159/commits/8272f88c3b50cc3912ae087b4f7eec506a5abaac). Made argument naming between `files` and `files_list` more coherent.

snekbox/nsjail.py
```diff
@@ -182,15 +182,12 @@ def python3(
         with NamedTemporaryFile() as nsj_log, MemFS(self.memfs_instance_size) as fs:
             # Add the temp dir to be mounted as cwd
             nsjail_args = (
+                # Set fslimit to unlimited, cannot be set in cfg
+                # due to upstream protobuf parsing issue
+                "--rlimit_fsize=inf",
```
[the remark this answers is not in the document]
ionite34: Thanks for that investigation. `rlimit_fsize_type: INF` works perfect.

[15e8df0](https://github.com/python-discord/snekbox/pull/159/commits/15e8df09356d53639dbe50034b82b1356dda6af8)

snekbox/memfs.py
ionite34: Yeah I agree, I think the context manager being optional is more in-line with conventions. I've changed it to be initialized during `__init__`, with a `weakref.finalize` running the cleanup on garbage collection with a `ResourceWarning` if the context manager is not used, modeled after https://github.com/python/cpython/blob/main/Lib/tempfile.py#L844

snekbox/api/resources/eval.py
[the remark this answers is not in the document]
ionite34: I've added compatible support for input in the API route in 84e73c1, so it is backwards compatible now. Added some associated unit tests to ensure the old `input` style still works.

Currently the API route requires either `args` or `input`. If `input` exists, `args` will have a default of `-c` if it is otherwise empty.

tests/test_memfs.py
```diff
@@ -0,0 +1,67 @@
...
+    def setUp(self):
+        super().setUp()
+        self.logger = logging.getLogger("snekbox.memfs")
+        self.logger.setLevel(logging.WARNING)
+        warnings.filterwarnings(
+            "ignore",
+            ".*Implicitly cleaning up.*",
+            ResourceWarning,
+            "snekbox.memfs",
+        )
```
MarkKoz, line 22: I think this filter may persist for other test cases. Is that a concern?

snekbox/memfs.py
```diff
@@ -0,0 +1,151 @@
...
+from snekbox.filesystem import mount, unmount
+from snekbox.snekio import FileAttachment
+
+log = logging.getLogger(__name__)
+
+__all__ = ("MemFS",)
+
+
+class MemFS:
+    """A memory temporary file system."""
```
MarkKoz, line 23:
```suggestion
    """An in-memory temporary file system."""
```

snekbox/memfs.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def __init__(self, instance_size: int, root_dir: str | Path = "/memfs") -> None:
+        """
+        Initialize a memory temporary file system.
```
MarkKoz, line 27:
```suggestion
        Initialize an in-memory temporary file system.
```

tests/api/test_eval.py
```diff
@@ -5,12 +5,16 @@ class TestEvalResource(SnekAPITestCase):
...
     def test_post_valid_200(self):
-        body = {"input": "foo"}
-        result = self.simulate_post(self.PATH, json=body)
-
-        self.assertEqual(result.status_code, 200)
-        self.assertEqual("output", result.json["stdout"])
-        self.assertEqual(0, result.json["returncode"])
+        cases = [
+            {"args": ["-c", "print('output')"]},
+            {"input": "print('hello')"},
```
MarkKoz, line 10: We should also test a case where both input and args are provided.

MarkKoz (changes requested):

MarkKoz:
Does the `output` directory have to be created beforehand or is that already being done somewhere?

snekbox/nsjail.py
```diff
@@ -35,11 +51,32 @@ def __init__(
...
+
+        Args:
+            nsjail_path: Path to the NsJail binary.
+            config_path: Path to the NsJail configuration file.
+            max_output_size: Maximum size of the output in bytes.
+            read_chunk_size: Size of the read buffer in bytes.
+            memfs_instance_size: Size of the tmpfs instance in bytes.
+            files_limit: Maximum number of output files to parse.
+            files_timeout: Maximum time in seconds to wait for output files to be read.
+            files_pattern: Pattern to match files to attach.
```
MarkKoz, line 70: Would it be possible to make the output directory part of the files_pattern? Would there be a problem with ambiguity in whether the directory should be created by NsJail or by the user?

README.md
```diff
@@ -60,10 +63,42 @@ The main features of the default configuration are:
 * Memory limit
 * Process count limit
 * No networking
-* Restricted, read-only filesystem
+* Restricted, read-only system filesystem
+* Memory-based virtual read-write filesystem mounted as cwd at `/home`
 
 NsJail is configured through [`snekbox.cfg`]. It contains the exact values for the items listed above. The configuration format is defined by a [protobuf file][7] which can be referred to for documentation. The command-line options of NsJail can also serve as documentation since they closely follow the config file format.
 
+### Memory File System
```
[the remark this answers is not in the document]
ionite34: Updated in [6ec4dd2](https://github.com/python-discord/snekbox/pull/159/commits/6ec4dd2d3026d1fe8e567b7eaf723d82811cdcf1)

snekbox/filesystem.py
```diff
@@ -0,0 +1,88 @@
...
+libc.umount2.argtypes = (ctypes.c_char_p, ctypes.c_int)
+
+
+class Size(IntEnum):
+    """Size multipliers for bytes."""
+
+    KB = 1024
+    MB = 1024**2
+    GB = 1024**3
+    TB = 1024**4
```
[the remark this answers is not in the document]
ionite34: Renamed in [4abc988](https://github.com/python-discord/snekbox/pull/159/commits/4abc988c6e5b9ede6cffe372d3252d9863c8eef0)

ionite34:
> Does the `output` directory have to be created beforehand or is that already being done somewhere?

It's done here 
https://github.com/python-discord/snekbox/blob/bytes-output/snekbox/memfs.py#L54

snekbox/nsjail.py
ionite34: Well currently we have the glob start at the output folder, otherwise we'd have the output folder in the glob result and have to strip it for sending.

If we allow the output to be in the glob, the folder will still be created by NSJail so an end user will still need to change the code to have a different output folder name for example.

README.md
```diff
@@ -60,10 +63,42 @@ The main features of the default configuration are:
...
+
+On each execution, the host will mount an instance-specific `tmpfs` directory,
+this is used as a limited read-write folder for the sandboxed code. There is no
+access of any kind to other files or directories on the host system or container,
+as only this new directory is mounted to NSJail.
```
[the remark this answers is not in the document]
ionite34: Yeah I think I need to rephrase this. Updated in [6ec4dd2](https://github.com/python-discord/snekbox/pull/159/commits/6ec4dd2d3026d1fe8e567b7eaf723d82811cdcf1)

README.md
```diff
@@ -60,10 +63,42 @@ The main features of the default configuration are:
...
+a visible folder `output`. The user has read/write access to any path under `home`.
+```
+/home
+ |- output
+```
+
+Files written to the `output` subfolder will be parsed and returned as
+a list of `FileAttachment` objects in `EvalResult.files` from the `python3` function.
+
+To send files to snekbox, it can be included as the `files` parameter of `python3`.
```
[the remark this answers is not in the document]
ionite34: Removed in [8c7da5a](https://github.com/python-discord/snekbox/pull/159/commits/8c7da5a32a6959409f917fb32ba331fe5132751e)

README.md
```diff
@@ -60,10 +63,42 @@ The main features of the default configuration are:
...
+
+Within the `/eval` route, files are attached or returned under the `files` key.
+
+> See [api/resources/eval.py](snekbox/api/resources/eval.py) for additional API schema information.
```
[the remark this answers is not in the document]
ionite34: Yup, removed in [5c588f7](https://github.com/python-discord/snekbox/pull/159/commits/5c588f7832eebfffc5886f25f48f10514b51a4fe)

tests/test_memfs.py
ionite34: This was due to the multithread test relying on implicit cleanup. I've since changed it to use contextlib.ExitStack instead to call `__exit__`, so this is not necessary and removed in [d33a55d](https://github.com/python-discord/snekbox/pull/159/commits/d33a55d775c9f15b6cd9089c7e2de3e827f0b837)

tests/api/test_eval.py
ionite34: Added in [95e3e18](https://github.com/python-discord/snekbox/pull/159/commits/95e3e18ac93709257840f8e5c36e6543678931f5)

snekbox/memfs.py
```diff
@@ -114,28 +89,46 @@ def mkdir(self, path: Path | str, chmod: int = 0o777) -> Path:
...
                 yield FileAttachment.from_path(file, relative_to=self.output)
 
+    def files_list(
+        self,
+        limit: int,
+        pattern: str,
+        preload_dict: bool = False,
+    ) -> list[FileAttachment]:
+        """
+        Returns a sorted list of output files found in the MemFS.
```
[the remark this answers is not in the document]
ionite34: Updated in [5d179ba](https://github.com/python-discord/snekbox/pull/159/commits/5d179ba5ee7e2d0c8808b8a7b90f6a560ab2acf0)

snekbox/snekio.py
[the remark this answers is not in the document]
ionite34: Updated regex to
```regex
^(?!/)(?!.*\\0).*$
```
This should disallow anything beginning with `/` or `\0` contained anywhere.

tests/test_libmount.py
```diff
@@ -1,17 +1,23 @@
...
-        super().setUp()
+    temp_dir = Path("/tmp/snekbox")
+
+    @classmethod
+    def setUpClass(cls):
+        cls.temp_dir.mkdir(exist_ok=True, parents=True)
+
+    @classmethod
+    def tearDownClass(cls):
+        rmtree(cls.temp_dir, ignore_errors=True)
```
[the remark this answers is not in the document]
ionite34: Actually speaking of `tempfile`, that's probably a better case, I've refactored this to use `tempfile.TemporaryDirectory` instead of manually doing this [62954cd](https://github.com/python-discord/snekbox/pull/159/commits/62954cd078cbcb4719cb1a1b726fe3672a17c495)

snekbox/nsjail.py
MarkKoz: I don't understand your answer, but I think it's because my question was unclear.

Currently, the output directory is hard-coded to `output`. I'm wondering if it'd be possible to make that configurable by allowing a user to specify the directory name in this `files_pattern` glob. However, I think this could result in ambiguity regarding what should be the path snekbox creates, and what should the end-user have to manually create. 

In any case, we can revisit it later. Not a blocker for the PR since I don't believe making this configurable in the future will be a breaking change.

snekbox/api/resources/eval.py
```diff
@@ -36,8 +36,8 @@ class EvalResource:
                     "properties": {
                         "path": {
                             "type": "string",
-                            # Disallow single absolute paths, and null bytes
-                            "pattern": r"^[^/\\0].*",
+                            # Disallow starting with / or containing \0 anywhere
+                            "pattern": r"^(?!/)(?!.*\\0).*$",
```
MarkKoz, line 40: Absolute case seems to be covered in tests but we do need to test the null byte as well.

README.md
```diff
@@ -60,10 +63,28 @@ The main features of the default configuration are:
...
 * No networking
-* Restricted, read-only filesystem
+* Restricted, read-only system filesystem
+* Memory-based virtual read-write filesystem mounted as cwd at `/home`
 
 NsJail is configured through [`snekbox.cfg`]. It contains the exact values for the items listed above. The configuration format is defined by a [protobuf file][7] which can be referred to for documentation. The command-line options of NsJail can also serve as documentation since they closely follow the config file format.
 
+### Memory File System
+
+On each execution, the host will mount an instance-specific `tmpfs` drive, this is used as a limited read-write folder for the sandboxed code. There is no access to other files or directories on the host container beyond the other read-only mounted system folders. Instance file systems are isolated, it is not possible for sandboxed code to access another instance's writeable directory.
```
MarkKoz, line 73: This is a comma splice. Use a semicolon or a full stop to create a separate sentence.
```suggestion
On each execution, the host will mount an instance-specific `tmpfs` drive, this is used as a limited read-write folder for the sandboxed code. There is no access to other files or directories on the host container beyond the other read-only mounted system folders. Instance file systems are isolated; it is not possible for sandboxed code to access another instance's writeable directory.
```

MarkKoz (changes requested):
Just a couple comments!

snekbox/nsjail.py
ionite34: Ah I see what you mean. I think implicitly creating paths in the glob might be a bit complex (it also would allow some arbitrary traversal and other situations) I've made the name of `home` and `output` configurable now in [c3a43ac](https://github.com/python-discord/snekbox/pull/159/commits/c3a43ac3b254efc999727c9fd15aac64d3203162)

So to not use output, `memfs_output=""` can be set to set the output glob root to home, or any other path.

snekbox/api/resources/eval.py
ionite34: Added additional tests in [55a9223](https://github.com/python-discord/snekbox/pull/159/commits/55a92231d83439b31441945799f713c25b569278) and checks to make sure separate `\` and `0` are allowed.

MarkKoz (approved):
!!! excellent work

MarkKoz:
@wookie184 Do you still plan on continuing your review?

ionite34:
Minor update in [3dfde52](https://github.com/python-discord/snekbox/pull/159/commits/3dfde52eda8f2af0475855c26bd5c1c2d2a0818d) to truncate `FileAttachment.__repr__` to avoid logs being spammed with megabytes of data from large user written files.

wookie184 (approved):
I don't think i'll have a chance to test this thoroughly, but this looks good.

snekbox/nsjail.py
```diff
@@ -163,37 +220,61 @@ def python3(
...
-                )
+                return EvalResult(args, None, "UnicodeDecodeError: invalid Unicode in output pipe")
 
             # When you send signal `N` to a subprocess to terminate it using Popen, it
             # will return `-N` as its exit code. As we normally get `N + 128` back, we
             # convert negative exit codes to the `N + 128` form.
             returncode = -nsjail.returncode + 128 if nsjail.returncode < 0 else nsjail.returncode
 
+            # Exclude user uploaded files from output
+            to_exclude = [file.path for file in files]
```
MarkKoz, line 262: Why send a string instead of a path object?

snekbox/nsjail.py
```diff
@@ -163,37 +220,61 @@ def python3(
...
-                    None,
-                )
+                return EvalResult(args, None, "UnicodeDecodeError: invalid Unicode in output pipe")
 
             # When you send signal `N` to a subprocess to terminate it using Popen, it
             # will return `-N` as its exit code. As we normally get `N + 128` back, we
             # convert negative exit codes to the `N + 128` form.
             returncode = -nsjail.returncode + 128 if nsjail.returncode < 0 else nsjail.returncode
 
+            # Exclude user uploaded files from output
```
MarkKoz, line 261: What if the files changed?

MarkKoz:
Is there a reason we would rather implement a file ignore system than disable Python bytecode generation? Are there eval use cases that would rely on it being generated?

Does the current ignore system support ignoring directories? The fnmatch docs state that a forward slash is not treated specially, but I'm unsure of what that means in practice. In any case, it looks more limited than gitignore so documentation is better off avoiding making such comparison.

ionite34:
> Is there a reason we would rather implement a file ignore system than disable Python bytecode generation? Are there eval use cases that would rely on it being generated?
> 
> Does the current ignore system support ignoring directories? The fnmatch docs state that a forward slash is not treated specially, but I'm unsure of what that means in practice. In any case, it looks more limited than gitignore so documentation is better off avoiding making such comparison.

it stemmed from this following code example which seems to generate a pyc file regardless of the PYTHONDONTWRITEBYTECODE env var on nsjail
```py
from pathlib import Path

Path("lib.py").write_text("func = lambda: 'hello'")

from lib import func

print(func())
```

snekbox/nsjail.py
ionite34: updated to upload user files if last modified time changes.
in  https://github.com/python-discord/snekbox/commit/36814de6f7a231c86b356ec27b649da5ffc85c44,

snekbox/nsjail.py
ionite34: is now a dictionary of Path and original last modified time floats in https://github.com/python-discord/snekbox/commit/36814de6f7a231c86b356ec27b649da5ffc85c44

MarkKoz:
Does using `python -B` fix it? https://docs.python.org/3/using/cmdline.html#cmdoption-B

ionite34:
> Does using `python -B` fix it? [docs.python.org/3/using/cmdline.html#cmdoption-B](https://docs.python.org/3/using/cmdline.html#cmdoption-B)

Seems to, yes. I removed the file ignore filter and added the -B flag.

MarkKoz (approved):
Thanks, glad that fixed it

PR #170, opened 2023-03-10
Exclude hidden paths in files output

ionite34:
To avoid cluttering files output with cache and configs from libraries like matplotlib, this will exclude parsing output files with a hidden (leading ".") directory or file name.

### DIFF
--- a/snekbox/memfs.py
+++ b/snekbox/memfs.py
@@ -138,6 +138,11 @@ class MemFS:
         """
         count = 0
         for file in self.output.rglob(pattern):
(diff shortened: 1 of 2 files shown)


MarkKoz:
Do we still ignore files starting with an underscore? Don't remember if it was implemented.

ionite34:
> Do we still ignore files starting with an underscore? Don't remember if it was implemented.

Yes, via the default nsjail / config `files_pattern: str = "**/[!_]*"`

PR #173, opened 2023-03-12
Fix recursion error during file attachment parsing of deep nested paths

ionite34:
Fixes: #172 

- Now catches the potential `RecursionError` from glob / pathlib. On such event we return an exit code `None`, and stdout:
```
FileParsingError: Exceeded directory depth limit while parsing attachments
```
- Additionally reduced file parsing time limit from 8 to 5 seconds.
- Added inter-loop time limit calculations within `MemFS.files` and `MemFS.files_list` loops with `time.monotonic` for cooperative time limit cancellation.
- Added a unit test with the reproducer from #172 

- Due to some weird interactions between `multiprocessing.Pool` and the mp runner for falcon resulting in a broken pipe and deadlock (see https://github.com/python-discord/snekbox/actions/runs/4395009986/jobs/7696475837), switched from mp time limit runner to a `SIGALRM` based one.


snekbox/nsjail.py
```diff
@@ -267,21 +267,32 @@ def python3(
...
+                    None,
+                    "FileParsingError: Exceeded directory depth limit while parsing attachments",
+                )
             except TimeoutError as e:
                 log.info(f"Exceeded time limit while parsing attachments: {e}")
                 return EvalResult(
                     args, None, "TimeoutError: Exceeded time limit while parsing attachments"
                 )
+            except Exception as e:
+                log.error(f"Unexpected {type(e).__name__} while parse attachments: {e}")
```
MarkKoz, line 292: Use `log.exception` so we get the traceback in logs, which I feel is important for a catch-all.

snekbox/nsjail.py
```diff
@@ -267,21 +267,32 @@ def python3(
...
-                attachments = timed(
-                    MemFS.files_list,
-                    (fs, self.files_limit, self.files_pattern),
-                    {
-                        "preload_dict": True,
-                        "exclude_files": files_written,
-                    },
-                    timeout=self.files_timeout,
-                )
+                with time_limit(self.files_timeout):
```
MarkKoz, line 270: Do we have a test that asserts this time outs?

snekbox/utils/timed.py
```diff
@@ -1,37 +1,34 @@
...
+    def signal_handler(_signum, _frame):
+        raise TimeoutError(f"time_limit call timed out after {timeout} seconds.")
+
+    signal.signal(signal.SIGALRM, signal_handler)
+    signal.alarm(timeout)
+
+    try:
+        yield
+    finally:
+        signal.alarm(0)
```
MarkKoz, line 34: Clever. I don't know enough about signals to say whether this is an appropriate use, but I think it's low risk to try.

snekbox/utils/timed.py
```diff
@@ -1,37 +1,34 @@
...
-    func: Callable[[_T], _V],
-    args: Iterable = (),
-    kwds: Mapping[str, Any] | None = None,
-    timeout: float | None = None,
-) -> _V:
+@contextmanager
+def time_limit(timeout: int | None = None) -> Generator[None, None, None]:
     """
-    Call a function with a time limit.
+    Decorator to call a function with a time limit. Uses SIGALRM, requires a UNIX system.
```
MarkKoz, line 16: I don't think implementation details should generally be included in the summary. If you think it's useful to point this out, then I think the notes section in the docstring is more appropriate of a place.

```suggestion
    Decorator to call a function with a time limit.
```

snekbox/memfs.py
```diff
@@ -176,15 +184,20 @@ def files_list(
                 Files will be excluded if their last modified time
                 is equal to the provided value.
             preload_dict: Whether to preload as_dict property data.
+            timeout: The maximum time for the file parsing. If exceeded,
+                a TimeoutError will be raised.
```
MarkKoz, line 188:
```suggestion
            timeout: The maximum time for the file parsing.
```
The omitted part should instead be in a `Raises:` section.

snekbox/memfs.py
```diff
@@ -135,12 +138,17 @@ def files(
             exclude_files: A dict of Paths and last modified times.
                 Files will be excluded if their last modified time
                 is equal to the provided value.
+            timeout: The maximum time for the file parsing. If exceeded,
+                a TimeoutError will be raised.
```
MarkKoz, line 142:
```suggestion
            timeout: The maximum time for the file parsing.
```
The omitted part should instead be in a `Raises:` section.

snekbox/memfs.py
```diff
@@ -135,12 +138,17 @@ def files(
...
         """
-        count = 0
-        for file in self.output.rglob(pattern):
-            # Ignore hidden directories or files
-            if any(part.startswith(".") for part in file.parts):
-                log.info(f"Skipping hidden path {file!s}")
+        start_time = time.monotonic()
+        added = 0
+        files = glob.iglob(pattern, root_dir=str(self.output), recursive=True, include_hidden=False)
+        for file in (Path(self.output, f) for f in files):
```
MarkKoz, line 147: Elegant use of a generator expression here!

snekbox/memfs.py
```diff
@@ -135,12 +138,17 @@ def files(
...
+            timeout: The maximum time for the file parsing. If exceeded,
+                a TimeoutError will be raised.
         """
-        count = 0
-        for file in self.output.rglob(pattern):
-            # Ignore hidden directories or files
-            if any(part.startswith(".") for part in file.parts):
-                log.info(f"Skipping hidden path {file!s}")
+        start_time = time.monotonic()
+        added = 0
```
MarkKoz, line 145: What's the thought behind renaming this to `added`? The previous name, `count`, was clearer to me. Nothing is being "added" (at least not as far as this function is concerned).

snekbox/memfs.py
```diff
@@ -150,21 +158,21 @@ def files(
...
+            log.info(f"Found valid file for upload {file.name!r}")
+            yield FileAttachment.from_path(file, relative_to=self.output)
 
     def files_list(
         self,
         limit: int,
         pattern: str,
         exclude_files: dict[Path, float] | None = None,
         preload_dict: bool = False,
+        timeout: float | None = None,
```
MarkKoz, line 175: Thoughts on using `float("inf")` (aka `math.inf`) instead of `None` to specify no timeout? Maybe a bit odd at first glance, but I think it makes sense, and it does avoid dealing with `None` checks.

MarkKoz (changes requested):

snekbox/memfs.py
ionite34: Yeah I think I'll change this back to count, previously I added then removed a `scanned` variable to stop at some determined depth hence leading to the name change.

ionite34: Changed back to `count` in [e924217](https://github.com/python-discord/snekbox/pull/173/commits/e9242175ced6584084b6a4a1c77d0319b49e2cc0)

snekbox/memfs.py
ionite34: Changed to `Raises:` in [6441719](https://github.com/python-discord/snekbox/pull/173/commits/64417197b58e3945f0b857c7fa648dcbe28254c4)

snekbox/memfs.py
ionite34: Changed to `Raises:` in [6441719](https://github.com/python-discord/snekbox/pull/173/commits/64417197b58e3945f0b857c7fa648dcbe28254c4)

snekbox/nsjail.py
ionite34: There is a test parsing symlink files for timeout, but that doesn't determine if the timeout occured due to the `time.monotic()` comparison timeout or by the signal. The signal timeout would be if a very large single file was parsed, or pathlib ends up in some loop. 

But also added 2 new tests for `time_limit` successfully interrupting `time.sleep` and a long-running built-in C function (math.factorial) in [1af5bf0](https://github.com/python-discord/snekbox/pull/173/commits/1af5bf0a92a5a65346460590054c1834ff7b7ccc)

snekbox/nsjail.py
ionite34: Changed to `log.exception` in [6b0dba5](https://github.com/python-discord/snekbox/pull/173/commits/6b0dba59e1d7a168e97531e5bd16ca04e5ed615a)

snekbox/memfs.py
ionite34: I feel like the None is still useful on performance grounds since we currently bypass the `time.monotic` call and comparison when `timeout` is None.

snekbox/utils/timed.py
ionite34: It seems to work pretty accurately for interrupting `time.sleep` and built-in C functions in these tests [1af5bf0](https://github.com/python-discord/snekbox/pull/173/commits/1af5bf0a92a5a65346460590054c1834ff7b7ccc), so it seems okay for our usecase, and avoids the potential broken pipes issue with multiprocessing.

snekbox/nsjail.py
```diff
@@ -289,7 +289,7 @@ def python3(
                     args, None, "TimeoutError: Exceeded time limit while parsing attachments"
                 )
             except Exception as e:
-                log.error(f"Unexpected {type(e).__name__} while parse attachments: {e}")
+                log.exception(f"Unexpected {type(e).__name__} while parse attachments", exc_info=e)
```
MarkKoz, line 292: `exc_info` is redundant when `log.exception` is used within an `except` block.

MarkKoz (approved):

ChrisLovering (approved):

PR #174, opened 2023-03-13
Update dependency versions

shenanigansd:
Update all dependencies
Explicitly updated Sentry to `1.16.0`


pyproject.toml
```diff
@@ -33,7 +33,7 @@ dependencies = [
 
 [project.optional-dependencies]
 gunicorn = ["gunicorn>=20.1"]  # Lowest which supports wsgi_app in config.
-sentry = ["sentry-sdk[falcon]>=1.5.4"]
+sentry = ["sentry-sdk[falcon]>=1.16.0"]
```
MarkKoz, line 36: Mind adding a comment explaining that we want this for Falcon 3.0 support and linking to https://github.com/getsentry/sentry-python/pull/1733 ?

MarkKoz (changes requested):
Comment in pyproject.toml for the falcon constraint can be removed now.

Also this can and should be removed https://github.com/python-discord/snekbox/blob/main/snekbox/utils/logging.py#L27-L28 I looked at the sentry PR and confirmed this warning will no longer be shown since it will try to import the new name first and only import the deprecated one as a fall back.

pyproject.toml
shenanigansd: a7a5ff8f7fc919727acffe084622f98036c2abf6

MarkKoz (approved):
Thanks for taking care of this!

PR #175, opened 2023-03-15
Support Multiple Python Versions

### LINKED ISSUE #158: Allow Evaluation Using Multiple Python Versions
## Description
Install multiple python versions within the container, and add an option in the API to select between them.

## Rationale
Currently, the bot supports evaluation of both 3.10 and 3.11 code, something I think should be kept and expanded in the future (to include major/important python versions). This is achieved by using the latest 3.10 image published here. This is fine for now, but it makes it harder to provide feature parity and avoid issues when users switch between modes, and it makes it difficult to patch and fix security issues.

## Implementation
The way I see us doing this is by building all the versions of python we want to support directly into the image, then aiming the jail at the right executable. This would simplify implementation by having one server, with one codebase, while allowing us to evaluate user code on any version.

## Specifics
Some more specific implementation details that can be up for debate:
1. The docker image will simply install whatever versions of python you pass to it in the build arg, to well defined locations
2. We'll build two images in CI, one with all versions we want supported, and one lightweight for dev
3. We can expose an endpoint to get all supported versions
4. We should investigate the best way to install all the versions to the docker image. I imagine the best solution is to just start from an image with all the required build-tools, install all the python versions, and use that as the base for the image. I found [this](https://hub.docker.com/r/fkrull/multi-python) docker image which has all the versions from 2.3-3.9 installed, and it comes out to a whooping... 600 MBs! I don't think this specific project would work for us since it hasn't been updated since 3.9, but it does make the idea seem feasiable.
5. Are the exposed packages going to be an issue, or can we simply install all the same versions to all python interpreters.

### LINKED ISSUE #168: Add Support For Running Other Python Implementations
Currently, the only Python implementation supported is the default CPython. Additional support should be added for executing Python code in other Python implementations like PyPy, RustPython, IronPython, etc. To support this, these implementations would simply have to be installed in the NSJail environment and alternate functions would have to be provided in the `NsJail` class for executing with alternate implementations.

HassanAbouelela:
Closes #158
Adds support for evaluating code in different python versions via the `version` argument.

This PR is a PoC to discuss the approach, and find issues with the idea. There's a versions.json file which specifies the supported python versions, and a make target for re-generating the relevant configuration files and dockerfile to enable the new versions. Packages are isolated based on python version.

The current approach was chosen for being relatively easy, quick, and lightweight when creating the container, however it is a little hacky. I am open to suggestions on other good methods for installing multiple python versions. Part of this proposal is to publish two different snekbox images, one with all images, and a lightweight one with just one version for developers.

Other things I've attempted included building all the versions in the container, manually and with pyenv, but those inevitably take way too long and produce really large containers. I've also tried adding deadsnakes and installing from apt, but I struggled to actually get it working within our image. This might still be a viable solution if we want to investigate it.


snekbox/api/resources/eval.py
```diff
@@ -54,6 +61,29 @@ class EvalResource:
     def __init__(self, nsjail: NsJail):
         self.nsjail = nsjail
 
+    @validate(
+        resp_schema={
+            "versions": {"type": "array", "items": {"type": "str"}},
+        }
+    )
+    def on_get(self, _: falcon.Request, resp: falcon.Response) -> None:
```
MarkKoz, line 69: Using GET on `/eval` to get this info seems unintuitive. I think this is better suited to be a dedicated "info" route. Related: #138

snekbox/api/resources/eval.py
```diff
@@ -54,6 +61,29 @@ class EvalResource:
...
+        """
+        Get information about the server.
+
+        Response format:
+        >>> {
+        ...     "versions": ["Python 3.9", "Python 3.10", "Python 3.12 Beta 1"]
```
MarkKoz, line 75: Maybe we should follow a naming convention more like pyenv, which distinguishes between cpython, pypy, etc.

snekbox/api/resources/eval.py
```diff
@@ -29,6 +32,10 @@ class EvalResource:
         "properties": {
             "input": {"type": "string"},
             "args": {"type": "array", "items": {"type": "string"}},
+            "version": {
+                "type": "string",
+                "oneOf": [{"const": name} for name in _VERSION_DISPLAY_NAMES],
+            },
```
MarkKoz, line 38: Alternative proposal: pass a path to an executable instead of a version. This would leave the door open for supporting arbitrary executables in the future. So far nothing in the API is actually specific to Python or even a compiler/interpreter. The downside is that it's less user-friendly to know the full path to an executable than it is to specify a Python version.

MarkKoz (commented):
Does this decouple the interpreter the web app uses and what eval/nsjail uses?

snekbox/api/resources/eval.py
HassanAbouelela: Sounds good, will change.

snekbox/api/resources/eval.py
HassanAbouelela: That sounds smart, should be fairly easy to modify in the json file.

HassanAbouelela:
It is sort of decoupled in the sense they are not enforced to be the same version as the one in the pyproject. The server will use one of the versions in versions.json, particularly the one labeled main.

MarkKoz:
I think the version the application uses should always match what we target with `pip-compile`. The frozen dependencies may not be compatible across Python versions, but this is not something I have looked into. 

I don't know if there is any way for us to actually enforce this. I suppose we can mostly rely on CI to tell us whether the dependencies are compatible.

snekbox/api/resources/eval.py
MarkKoz: So instead of discovering which versions are available, users could discover which binaries are available and what their paths are. We don't need to list _everything_ that's mounted; can continue to use the same approach of configuring that in a JSON file.

HassanAbouelela:
The main challenge is figuring that version out during the dockerfile creation step, and verifying the version string against whatever is configured in the file. It might be doable, do we want that enforcement?

MarkKoz:
No, I don't think we need to do anything for that. I think the tests + image build than run in CI gives us confidence in the dependencies being compatible. It's not worth adding complexity to our image. If there was a straightforward way, then sure.

snekbox/api/resources/eval.py
HassanAbouelela: This is an interesting idea, I'll into implementing it. I'd probably keep the display names in the JSON files so end-users (bot) can still display human-readable names.

MarkKoz: Another approach would be to base it off nsjail config files. Each binary (including different Python versions) would have its own config file, and we would scan a directory for config files. The API would accept a config name. This would offer more flexibility in supporting multiple binaries, as different binaries may require different mounts, environment variables, etc. I don't think this approach is mutually exclusive with my previous proposal, so we can revisit this in the future as well.

HassanAbouelela: I wonder how much of the config will be repeated. I imagine a lot of the python (for the cpython implementations) configs will be very repetitive, but even other interpreters would presumably need similar cgroup configuration, mounts, etc. We might be able to get around the duplication problem by providing overrides to be applied in python (or if nsjail supports combining configs?) instead of multiple config files. Either way, it's a problem for another day.

snekbox/api/resources/eval.py
HassanAbouelela: Done in 8375787. I didn't implement the rest of the suggestions from that issue, but I'd think they'd be easy to implement in a future PR.

snekbox/api/resources/eval.py
HassanAbouelela: f564fd70

snekbox/api/resources/eval.py
HassanAbouelela: I'm reviewing this again - what's the benefit to passing binary paths instead of version names. Persumably we're not going to take in arbitrary paths, but one from a list of pre-defined ones such as the ones in version.json. That allows us to continue giving them human-readable names. At that point the path is an implementation detail with end users only caring about the name (CPython x.y.z vs pypy x.y.z for example).

HassanAbouelela:
Nearly done, just need to wrap up some dependency work. I've tested the new CI on [my fork](https://ghcr.io/hassanabouelela/snekbox).

snekbox/api/resources/eval.py
MarkKoz: Yeah, fair enough if we stick with something like versions.json. However, I am considering that maybe arbitrary paths is better. It's more flexible, it's less complex, and requires no additional configuration. What problem or use case are we trying to solve by restricting the binaries to a predefined set?

HassanAbouelela: We'd need to define the python versions statically somewhere at any rate, so we can actually define them in the dockerfile. We might be able to get away with an image that has all python versions already so we don't need to specify them, but that would require an external image most likely. I linked an example in the original issue, but that is no longer maintained.

On a more general level though, I'm not sure arbitrary binaries is the best approach. Different binaries would require different configurations, be that resources, mounts, or permissions. The most flexible system would be statically defined sets of configurations - harking back to your multiple configs idea - and at that point we'd end up with end users selecting from a pre-defined set of human names.

This system could actually define multiple binaries per config-set as well by passing the path as an argument, or more ideally, defining the paths in a dict and passing the name to the API which resolves the correct binary.

MarkKoz: > The most flexible system would be statically defined sets of configurations

I agree, and we want to leave the door open for that in the future.

> Different binaries would require different configurations, be that resources, mounts, or permissions. 

This is true, but not always. It is feasible for a single NsJail config to support multiple binaries (though it can only have one default entry point I think).

> We'd need to define the python versions statically somewhere at any rate

That only applies to Python, and I consider it an implementation detail as far as the API is concerned. I believe we need a stronger reason to influence the design of our API.

> defining the paths in a dict and passing the name to the API which resolves the correct binary.

I find this redundant. It seems simpler and more intuitive to create an NsJail config, which effectively defines an environment, and have everything in that environment be available via the API.

HassanAbouelela: > It is feasible for a single NsJail config to support multiple binaries (though it can only have one default entry point I think).

This is true, but we can probably just avoid defining a default entry point in the first place. That's more or less what this PR does for python.

> It seems simpler and more intuitive to create an NsJail config, which effectively defines an environment, and have everything in that environment be available via the API.

My problem is it moves the burden of knowing what binaries are available on the API from the project (which can reasonably control the matter) to the user (which can not directly affect it). For instance, if we use this PR to implement support for multiple python versions on bot, we can't trivially define what versions are available to the bot without essentially hardcoding it. A change to snekbox would necessitate a change to the users as well.

Regardless of which approach we choose, it'd be our burden as maintainers to ensure if we provide a binary for eval in one version, that we also have it available for future versions, or declaring an incompatibility if needed. This is much harder when we don't actually know what binaries users need, and have to assume everything is used.

Explicitly defining which binaries the API supports (and returning them in `/info`) alleviates both problem. At that point a display name is just the icing on the cake.

ChrisLovering:
Closing this now that https://github.com/python-discord/snekbox/pull/181 is merged

PR #185, opened 2023-08-29
Refactor how we process eval dependencies

jb3:
Right now, our eval dependencies are pulled in using a script embedded within the deployment file, and all dependencies are listed there. This gets a bit messy the more dependencies we want to add, especially if we want to add version markers.

This PR splits out the dependencies into a new file under `requirements/` and moves the dependency install process to it's own init container instead of being a task that executes within the snekbox container after it starts. This also allows for greater observability into the dependencies pull process as logs can be fetched from init containers with `kubectl logs deploy/snekbox -c deps-install`.

### DIFF
--- a/deployment.yaml
+++ b/deployment.yaml
@@ -12,6 +12,22 @@ spec:
       labels:
         app: snekbox
     spec:
+      initContainers:
+        - name: deps-install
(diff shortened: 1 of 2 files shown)


ChrisLovering (approved):

jb3:
Thanks!

PR #190, opened 2023-08-30
Make CI and deployment.yaml share the same dependency install script

ChrisLovering:


scripts/install_eval_deps.sh
```diff
@@ -0,0 +1,3 @@
+export PYTHONUSERBASE=/snekbox/user_base
```
MarkKoz, line 1: Would be good to add `set -euo pipefail` as sane defaults in case anything ever changes in this script

MarkKoz (changes requested):

scripts/install_eval_deps.sh
ChrisLovering: Added in https://github.com/python-discord/snekbox/pull/190/commits/b0180ee7cded3ca51ab832892ae162f13a0e1783 had to change it to run under bash for `-o pipefail` support.

I've also added a make target since I found myself copy-pasting from CI a lot while testing this.

MarkKoz (approved):
Thanks

PR #193, opened 2023-09-16
Update to Debian Bookworm

MarkKoz:
Mount /dev/shm in the container by setting ipc to "private". This is the same as "none" (the previous value) with the only difference being that shm is mounted. This is needed for integration tests to pass.

The integration tests always relied on shared memory due to their use of multiprocessing. They managed to work because glibc used to fall back to /tmp if /dev/shm wasn't available. However, newer versions of glibc, which Debian Bookworm now uses, removed that fallback behaviour.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # syntax=docker/dockerfile:1.4
-FROM buildpack-deps:buster as builder-nsjail
+FROM buildpack-deps:bookworm as builder-nsjail
 
 WORKDIR /nsjail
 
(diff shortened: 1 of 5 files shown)


ChrisLovering (changes requested):
Runs fine. Tested with it integrated tot he bot, works fine too. However ruining the tests locally, I get this failure.

```
FAIL: test_file_parsing_timeout (tests.test_nsjail.NsJailTests.test_file_parsing_timeout)
----------------------------------------------------------------------
  File "/snekbox/tests/test_nsjail.py", line 247, in test_file_parsing_timeout
    self.assertEqual(result.returncode, None)
AssertionError: 0 != None
```
Running the code from that test against the bot seems like snekbox returns just fine and it (expectedly) hits errors in the bot's callback
image

ChrisLovering (approved):
Works well, thanks Mark!

PR #196, opened 2023-10-10
Multiple binary support api

ChrisLovering:
This allows setting a `binary_path` key in the request body to `/eval` to specify which binary to run under.

If not supplied, it defaults to `"/snekbin/python/default/bin/python"`.

https://github.com/python-discord/snekbox/pull/196/commits/0c9b234011f21ce7bdc993957867610dfec1258f is needed as previously we were dropping the text after the last period when creating the path to install Python to. However, for versions such as3.13-dev this would result in it being installed into `/snekbin/3`. Instead, we now split on the last `.` _or_ the last `-`

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,9 +52,9 @@ The above command will make the API accessible on the host via `http://localhost
 
 ### Python multi-version support
 
-By default, the binary that runs within nsjail is the binary specified by `DEFAULT_BINARY_PATH` at the top of [`nsjail.py`]. This can be overridden by specifying `binary_path` in the request body of calls to `POST /eval` or by setting the `binary_path` kwarg if calling `NSJail.python3()` directly.
+By default, the executable that runs within nsjail is defined by `DEFAULT_EXECUTABLE_PATH` at the top of [`nsjail.py`]. This can be overridden by specifying `executable_path` in the request body of calls to `POST /eval` or by setting the `executable_path` kwarg if calling `NSJail.python3()` directly.
 
-Any binary that exists within the container is a valid value for `binary_path`. The main use case of this feature is currently to specify the version of Python to use.
+Any executable that exists within the container is a valid value for `executable_path`. The main use case of this feature is currently to specify the version of Python to use.
 
 Python versions currently available can be found in the [`Dockerfile`] by looking for build stages that match `builder-py-*`. These binaries are then copied into the `base` build stage further down.
 
--- a/README.md
+++ b/README.md
@@ -50,6 +50,14 @@ To run it in the background, use the `-d` option. See the documentation on [`doc
 
 The above command will make the API accessible on the host via `http://localhost:8060/`. Currently, there's only one endpoint: `http://localhost:8060/eval`.
 
+### Python multi-version support
+
+By default, the binary ran within nsjail is the binary specified by `DEFAULT_BINARY_PATH` at the top of [`nsjail.py`]. This can be overridden by specifying `binary_path` in the request body of calls to `POST /eval` or by setting the `binary_path` kwarg if calling `NSJail.python3()` directly.
+
+Any binary that exists within the container is a valid value for `binary_path`. The main use case of this feature is currently to specify the version of Python to use.
+
+Python versions currently available can be found in the [`Dockerfile`] by looking for build stages that match `builder-py-*`. These binaries are then copied into the `base` build stage further down.
+
 ## Configuration
 
 Configuration files can be edited directly. However, this requires rebuilding the image. Alternatively, a Docker volume or bind mounts can be used to override the configuration files at their default locations.
@@ -126,9 +134,11 @@ See [CONTRIBUTING.md](.github/CONTRIBUTING.md).
 [7]: https://github.com/google/nsjail/blob/master/config.proto
 [`gunicorn.conf.py`]: config/gunicorn.conf.py
 [`snekbox.cfg`]: config/snekbox.cfg
+[`nsjail.py`]: snekbox/nsjail.py
 [`snekapi.py`]: snekbox/api/snekapi.py
 [`resources`]: snekbox/api/resources
 [`docker-compose.yml`]: docker-compose.yml
+[`Dockerfile`]: Dockerfile
 [`docker run`]: https://docs.docker.com/engine/reference/commandline/run/
 [nsjail]: https://github.com/google/nsjail
 [falcon]: https://falconframework.org/
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ To expose third-party Python packages during evaluation, install them to a custo
 
 ```sh
 docker exec snekbox /bin/sh -c \
-    'PYTHONUSERBASE=/snekbox/user_base /lang/python/default/bin/python -m pip install --user numpy'
+    'PYTHONUSERBASE=/snekbox/user_base /snekbin/python/default/bin/python -m pip install --user numpy'
 ```
 
 In the above command, `snekbox` is the name of the running container. The name may be different and can be checked with `docker ps`.
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ The above command will make the API accessible on the host via `http://localhost
 
 ### Python multi-version support
 
-By default, the binary ran within nsjail is the binary specified by `DEFAULT_BINARY_PATH` at the top of [`nsjail.py`]. This can be overridden by specifying `binary_path` in the request body of calls to `POST /eval` or by setting the `binary_path` kwarg if calling `NSJail.python3()` directly.
+By default, the binary that runs within nsjail is the binary specified by `DEFAULT_BINARY_PATH` at the top of [`nsjail.py`]. This can be overridden by specifying `binary_path` in the request body of calls to `POST /eval` or by setting the `binary_path` kwarg if calling `NSJail.python3()` directly.
 
 Any binary that exists within the container is a valid value for `binary_path`. The main use case of this feature is currently to specify the version of Python to use.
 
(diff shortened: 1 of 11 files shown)


snekbox/nsjail.py
```diff
@@ -188,7 +189,12 @@ def python3(
             py_args: Arguments to pass to Python.
             files: FileAttachments to write to the sandbox prior to running Python.
             nsjail_args: Overrides for the NsJail configuration.
+            binary_path: The path to the binary to execute under.
+                If None, defaults to "/lang/python/default/bin/python"
```
MarkKoz, line 193: I think we should rename `/lang` since it's going to be more prominent now. Maybe `/snekbin`? — doubles as a kinda clever play on "snekbox". If so, we also have to rename our references to the path in our CONTRIBUTING.md docs.

tests/test_integration.py
```diff
@@ -52,6 +52,54 @@ def test_memory_limit_separate_per_process(self):
...
+                with self.subTest(msg=msg, body=body, expected=expected):
+                    response, status = snekbox_request(body)
+                    self.assertEqual(status, 200)
+                    self.assertEqual(json.loads(response)["stdout"], expected)
+
+    def invalid_binary_paths(self):
+        """Test that passing invalid binary paths result in no code execution."""
+        with run_gunicorn():
+            cases = [
+                ("/bin/bash", "test files outside of /lang cannot be ran"),
```
MarkKoz, line 88: typo
```suggestion
                ("/bin/bash", "test files outside of /lang cannot be run"),
```

tests/test_integration.py
```diff
@@ -52,6 +52,54 @@ def test_memory_limit_separate_per_process(self):
...
+                (
+                    "/lang/../bin/bash",
+                    "test path traversal still stops files outside /lang from running",
+                ),
+                ("/foo/bar", "test non-existant files are not ran"),
```
MarkKoz, line 93: typo
```suggestion
                ("/foo/bar", "test non-existant files are not run"),
```

MarkKoz (changes requested):
Should we add something in the readme about multiple version support? Can just be a blurb I think.

snekbox/api/resources/eval.py
```diff
@@ -123,10 +125,21 @@ def on_post(self, req: falcon.Request, resp: falcon.Response) -> None:
         if "input" in body:
             body.setdefault("args", ["-c"])
             body["args"].append(body["input"])
+
+        binary_path = body.get("binary_path")
+        if binary_path:
+            binary_path = Path(binary_path)
+            if (
+                not binary_path.resolve().as_posix().startswith("/lang/")
```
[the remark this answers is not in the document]
MarkKoz: Oh I missed this in your description

> if supplied, snekbox ensures that the path is within /lang (to avoid executing arbitrary binaries on the fs) and that the file exists.

It is fairly trivial to bypass this restriction by using the `execve` syscall. We cannot block this syscall either (at least not through nsjail's seccomp filters). That is due to a quirk in which the filter is applied before nsjail's own `execve` call, causing nsjail to commit suicide.

Granted, there may be use cases which don't have binaries in `/lang` that make it possible to call syscalls. But I think if a request comes up to have such restriction, we can add it as a later feature.

ChrisLovering:
I plan to pick this PR up soon just had time to rebase this evening

snekbox/api/resources/eval.py
```diff
@@ -122,10 +124,21 @@ def on_post(self, req: falcon.Request, resp: falcon.Response) -> None:
...
+                or not binary_path.is_file()
```
jchristgit, line 133: Should we also check that it's executable?
```suggestion
                or not binary_path.is_file()
                or not binary_path.stat().st_mode & 0o100 == 0o100
```

snekbox/api/resources/eval.py
```diff
@@ -122,10 +124,21 @@ def on_post(self, req: falcon.Request, resp: falcon.Response) -> None:
         if "input" in body:
             body.setdefault("args", ["-c"])
             body["args"].append(body["input"])
+
+        binary_path = body.get("binary_path")
+        if binary_path:
+            binary_path = Path(binary_path)
+            if (
+                not binary_path.resolve().as_posix().startswith("/lang/")
```
jchristgit, line 132: Should we move the `resolve().as_posix()` calls above so the pathlib calls below can make use of it? Or do they mutate the internal object (source of eldritch horrors)

jchristgit (commented):
This is really cool!!!

HassanAbouelela:
Hey how's this PR going? It would be very useful for a project I'm working on.

snekbox/api/resources/eval.py
```diff
@@ -123,10 +125,21 @@ def on_post(self, req: falcon.Request, resp: falcon.Response) -> None:
...
+                or not binary_path.is_file()
+            ):
+                raise falcon.HTTPBadRequest(title="binary_path file is invalid")
```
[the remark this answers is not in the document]
ChrisLovering: https://github.com/python-discord/snekbox/pull/196/commits/d936970e5f6460e4dc28a17b2a332211f5a6d850

snekbox/nsjail.py
```diff
@@ -188,7 +189,12 @@ def python3(
...
         """
+        if not binary_path:
+            binary_path = "/lang/python/default/bin/python"
```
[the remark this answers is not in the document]
ChrisLovering: Had to do something slightly idfferent so that hte default cna be used both by calling python3 directly, and when calling the API https://github.com/python-discord/snekbox/pull/196/commits/66c6d27335a5d1fc23bd0889d299cb79f3418219

snekbox/nsjail.py
ChrisLovering: Yeah I agree this is a good change https://github.com/python-discord/snekbox/pull/196/commits/f726695b1a1d99b5c61ea09bd3a364103a08d7bf

tests/test_integration.py
ChrisLovering: https://github.com/python-discord/snekbox/pull/196/commits/30d51d19ed6eb425cbfb85c5dcbebd5e77816732

tests/test_integration.py
ChrisLovering: https://github.com/python-discord/snekbox/pull/196/commits/30d51d19ed6eb425cbfb85c5dcbebd5e77816732

snekbox/api/resources/eval.py
ChrisLovering: Yes! https://github.com/python-discord/snekbox/pull/196/commits/81573b2d0144ece5490948706918233b887f86c1

snekbox/api/resources/eval.py
ChrisLovering: changed this func quite a bit since then, so this is no longer relevant

ChrisLovering:
Now ready for (re)review

README.md
```diff
@@ -50,6 +50,14 @@ To run it in the background, use the `-d` option. See the documentation on [`doc
 
 The above command will make the API accessible on the host via `http://localhost:8060/`. Currently, there's only one endpoint: `http://localhost:8060/eval`.
 
+### Python multi-version support
+
+By default, the binary ran within nsjail is the binary specified by `DEFAULT_BINARY_PATH` at the top of [`nsjail.py`]. This can be overridden by specifying `binary_path` in the request body of calls to `POST /eval` or by setting the `binary_path` kwarg if calling `NSJail.python3()` directly.
```
MarkKoz, line 55:
```suggestion
By default, the binary that runs within nsjail is the binary specified by `DEFAULT_BINARY_PATH` at the top of [`nsjail.py`]. This can be overridden by specifying `binary_path` in the request body of calls to `POST /eval` or by setting the `binary_path` kwarg if calling `NSJail.python3()` directly.
```

tests/test_integration.py
```diff
@@ -52,6 +52,54 @@ def test_memory_limit_separate_per_process(self):
...
+                    response, status = snekbox_request(body)
+                    self.assertEqual(status, 200)
+                    self.assertEqual(json.loads(response)["stdout"], expected)
+
+    def invalid_binary_paths(self):
+        """Test that passing invalid binary paths result in no code execution."""
+        with run_gunicorn():
+            cases = [
+                ("/abc/def", "test non-existant files are not run", "binary_path does not exist"),
+                ("/snekbin", "test directories are not ran", "binary_path is not a file"),
```
MarkKoz, line 89:
```suggestion
                ("/abc/def", "test non-existent files are not run", "binary_path does not exist"),
                ("/snekbin", "test directories are not run", "binary_path is not a file"),
```

tests/test_integration.py
```diff
@@ -52,6 +52,54 @@ def test_memory_limit_separate_per_process(self):
...
+                (
+                    "/etc/hostname",
+                    "test non-executable files are not ran",
```
MarkKoz, line 92:
```suggestion
                    "test non-executable files are not run",
```

snekbox/api/resources/eval.py
```diff
@@ -43,6 +44,7 @@ class EvalResource:
                     "required": ["path"],
                 },
             },
+            "binary_path": {"type": "string"},
```
MarkKoz, line 47: Maybe "executable_path" is a better name than "binary_path"? I'm also okay with the latter though.

MarkKoz (changes requested):

snekbox/api/resources/eval.py
ChrisLovering: Yeah I agree, makes it clearer what is supported here, as it's not limitted to just binaries. https://github.com/python-discord/snekbox/pull/196/commits/edcaed6ad90f71ff61f6ceb7194de611ceeec592

MarkKoz (approved):
thanks!

PR #198, opened 2024-01-24
Bump ci

ChrisLovering:
This bumps various actions in CI, and upgrades the pre-commit version used locally.


wookie184 (approved):
All looks good 👍

PR #200, opened 2024-03-14
Bump snekbox up to Python 3.12.2

ChrisLovering:
pyenv 2.3.36 added support for 3.12.2, it also happens to be the latest version as of writing.

I also bumped eval-deps up and reenabled yarl as multidict has 3.12 wheels now.


shtlrs (approved):
Untested, otherwise looks good to me

PR #208, opened 2024-04-03
Use xargs over find -exec to ensure exit code is returned properly

ChrisLovering:
Running this script in it's previous form (via `docker compose run`) always returned an exit code of 0. This is due to `find` always returning a 0 exit code, unless an error occurred while traversing the directories.

This caused CI to always report a success when isntalling deps, even when pip failed.


ChrisLovering:
forcepush was to reword the commit description to be clearer.

jchristgit (approved):
**Great!**

MarkKoz (approved):
Nice, thanks!

Issue #205, opened 2024-03-15
Weird-byte filenames trigger internal server error on json encode


bast0006:
The following test code:

```py
with open(b"\xC3.txt", "w") as f: 
  f.write("test")
```

Snekbox fails with a (presumably non-exploitable) 500 error:

```
[FALCON] [ERROR] POST /eval => Traceback (most recent call last):
File "falcon/app.py", line 386, in falcon.app.App.__call__
File "falcon/app.py", line 1070, in falcon.app.App._get_body
File "falcon/response.py", line 270, in falcon.response.Response.render_body
File "falcon/media/json.py", line 179, in falcon.media.json.JSONHandler._serialize_s	
UnicodeEncodeError: 'utf-8' codec can't encode character '\udcc3' in position 279: surrogates not allowed
```

This is because the filename for included files (since https://github.com/python-discord/bot/pull/2326 ) doesn't handle arbitrary bytes in filenames. 

Presumably safe_path (https://github.com/python-discord/snekbox/blob/d6e940065552d2e46a0ddebd72dcd9b6a8348c4a/snekbox/snekio/attachment.py#L14) needs to be updated to catch invalid byte sequences that will fail JSON encoding.

Esther-Goldberg:
Can I work on this?

MarkKoz:
> Can I work on this? 
> 

Yes!

PR #209, opened 2024-04-07
raise error if attachment pathname contains invalid bytes

Esther-Goldberg:
safe_path is used for input, to parse a file path from the dict. This error is occurring when doing the reverse. This change throws an EvalError if any file created by the Python code isn't able to be encoded in UTF-8.


Esther-Goldberg:
This would close #205

Esther-Goldberg:
Sure! I'll add that once I get it done

Esther-Goldberg:
Is a test for the error returned by nsjail enough, or should the function that creates files from paths also have a test?

MarkKoz:
That test is sufficient - it's what I also had in mind. However, the test is failing in CI.

MarkKoz (approved):

PR #217, opened 2024-10-03
Py 3.13 rc

ChrisLovering:
This adds support for Python 3.13 to snekbox.

Actually calling this isn't possible until https://github.com/python-discord/snekbox/pull/196 is merged, which I will be looking into now.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -52,6 +52,7 @@ jobs:
           name: coverage-${{ matrix.os }}
           path: .coverage.*
           retention-days: 1
+          include-hidden-files: true
 
   report:
     name: Report coverage
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,13 +29,16 @@ RUN apt-get -y update \
         tk-dev \
     && rm -rf /var/lib/apt/lists/*
 
-RUN git clone -b v2.4.10 --depth 1 https://github.com/pyenv/pyenv.git $PYENV_ROOT
+RUN git clone -b v2.4.14 --depth 1 https://github.com/pyenv/pyenv.git $PYENV_ROOT
 
 COPY --link scripts/build_python.sh /
 
 # ------------------------------------------------------------------------------
 FROM builder-py-base as builder-py-3_12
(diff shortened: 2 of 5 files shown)


.github/workflows/test.yaml
```diff
@@ -45,6 +45,9 @@ jobs:
             snekbox \
             run -m unittest
 
+      - name: debug fs
```
MarkKoz: Did you intended on leaving this in?

Dockerfile
```diff
@@ -29,12 +29,16 @@ RUN apt-get -y update \
         tk-dev \
     && rm -rf /var/lib/apt/lists/*
 
+RUN git clone -b v2.4.14 --depth 1 https://github.com/pyenv/pyenv.git $PYENV_ROOT
```
MarkKoz, line 32: Excellent change!

snekbox/__main__.py
```diff
@@ -13,10 +13,16 @@ def parse_args() -> argparse.Namespace:
     )
     parser.add_argument("code", help="the Python code to evaluate")
     parser.add_argument(
-        "nsjail_args", nargs="?", default=[], help="override configured NsJail options"
+        "nsjail_args",
```
MarkKoz, line 16: Do you have a sample of what the help output is after this change?

.github/workflows/test.yaml
ChrisLovering: lol no

snekbox/__main__.py
ChrisLovering: Yea, the reason I added the default to the help directly is because it didn't get added automatically.

```
❯ python3 -m snekbox --help
usage: snekbox [-h] code [nsjail_args ...] [--- py_args ...]

positional arguments:
  code         the Python code to evaluate
  nsjail_args  override configured NsJail options (default: [])
  py_args      arguments to pass to the Python process (default: ['-c'])

options:
  -h, --help   show this help message and exit
```

MarkKoz: Thanks - that still looks helpful

MarkKoz (approved):

PR #236, opened 2025-05-17
Python 3.14

ChrisLovering:
This should be merged with the linked [bot PR](https://github.com/python-discord/bot/pull/3324)

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -11,11 +11,7 @@ on:
 jobs:
   test:
     name: Test with coverage
-    runs-on: ${{ matrix.os }}
-    strategy:
-      matrix:
-        os: [ubuntu-22.04, ubuntu-24.04]
-
+    runs-on: ubuntu-latest
     steps:
       - name: Download image artifact
         uses: actions/download-artifact@v4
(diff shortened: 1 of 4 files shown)


jb3 (approved):

GriceTurrble (approved):

.github/workflows/test.yaml
```diff
@@ -14,7 +14,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-20.04, ubuntu-22.04]
+        os: [ubuntu-22.04, ubuntu-24.04]
```
MarkKoz, line 17: I think we were using Ubuntu 20 so we could test cgroups v1

ChrisLovering: Ubuntu 20.04 stopped being available last month, so we can't use that runner anymore https://github.com/actions/runner-images?tab=readme-ov-file#available-images

MarkKoz: Oh well. I suppose we could set up a custom runner but I don't think that's for it just for v1 coverage.

ChrisLovering: Yeah probably not worth the headache of setting up, but sad it'll be untested now though. In that case I'll just update this CI to run on ubuntu-latest instead

ChrisLovering: https://github.com/python-discord/snekbox/pull/236/commits/3027f4b9cecab1f3f4e05eeefdcbd33593e54745

PR #237, opened 2025-05-31
Pin 3.14 dev version

ChrisLovering:


Makefile
```diff
@@ -28,7 +28,7 @@ lint: setup
 test:
 	docker compose build -q --force-rm
 	docker compose run --entrypoint /bin/bash --rm snekbox -c \
-    	'coverage run -m unittest; e=$?; chown --reference=. .coverage; exit $e'
+		'coverage run -m unittest; e=$?; chown --reference=. .coverage; exit $e'
```
ChrisLovering, line 31: this is swapping 4 spaces to a tab, as the rest of the file is all tabs

Dockerfile
```diff
@@ -41,7 +41,7 @@ FROM builder-py-base AS builder-py-3_13t
 RUN /build_python.sh 3.13.2t
 # ------------------------------------------------------------------------------
 FROM builder-py-base AS builder-py-3_14
-RUN /build_python.sh 3.14-dev
+RUN /build_python.sh 3.14.0b2
```
ChrisLovering, line 44: patch ver & extras are ignored by the install script, so this doesn't change the file path that it is installed to, which is `/snekbin/python/3.14/`

PR #238, opened 2025-07-30
Upgrade Python versions & eval-deps

ChrisLovering:


tests/test_nsjail.py
```diff
@@ -458,6 +458,8 @@ def test_multiprocessing_shared_memory(self):
 
                     shm = SharedMemory(create=True, size={shm_size})
                     shm.buf[:{buffer_size}] = bytearray([1] * {buffer_size})
+                    shm.close()
+                    shm.unlink()
```
jb3, line 462: Why did this become necessary?

jb3: Or rather, was this because of the Python version upgrade?

ChrisLovering: Looks related to the new version yeah. It's from the [resource_tracker](https://github.com/python/cpython/blob/9d3b53c47fab9ebf1f40d6f21b7d1ad391c14cd7/Lib/multiprocessing/resource_tracker.py#L297-L301), which the [docs say](https://docs.python.org/3/library/multiprocessing.shared_memory.html#multiprocessing.shared_memory.SharedMemory) (see track kwarg docs) was added in 3.13.

PR #241, opened 2025-08-24
Update all the versions

ChrisLovering:
Tests all _just passed_ locally. The nsjail bump does take us through a [major version](https://github.com/google/nsjail/releases/tag/3.0) update, but doesn't seem breaking for us.


jb3 (approved):
Chris, this is some of your greatest work yet. Keep up the good work! :shipit:

Dockerfile
```diff
@@ -1,23 +1,22 @@
...
+        bison \
         flex \
-        libprotobuf-dev\
+        libprotobuf-dev \
         libnl-route-3-dev \
         protobuf-compiler \
     && rm -rf /var/lib/apt/lists/*
 
-RUN git clone -b master --single-branch https://github.com/google/nsjail.git . \
-    && git checkout dccf911fd2659e7b08ce9507c25b2b38ec2c5800
```
MarkKoz, line 16: I think we used a commit instead of a branch to minimise the amount of data we pull from the remote.

ChrisLovering: Wouldn't doing `-b master--single-branch` clone the entire master branch and then after that we checkout to a older commit.

Whereas the `-b 3.4 --depth 1` only clones the single commit that is tagged with that name?

happy to change it back if I've misunderstood that

MarkKoz: Sorry, I had it backwards. We originally had it as you wrote it now. It was changed to a commit because we wanted a specific unreleased version. 

See the commit below and its associated PR description. https://github.com/python-discord/snekbox/commit/b1bcd3b04fb2a83fd7bd4fe930582ccf0b6bf627

ChrisLovering: Ah makes sense!

MarkKoz:
When we update nsjail we should regenerate the protobuf code for its config. I believe our README explains this process.

ChrisLovering:
> When we update nsjail we should regenerate the protobuf code for its config. I believe our README explains this process.

Ah cool, didn't know about this, I'm assuming it's `./scripts/protoc.py 3.4` can look tomorrow unless joe gets it first

MarkKoz (approved):

PR #248, opened 2025-10-18
Separate and simplify PyDis specific usage

jb3:
This PR makes a start at decoupling the specific setup of snekbox PyDis uses and snekbox as a standalone application.

This is done by building a basic snekbox application that supports evaluation using only the system version of Python.

We then have a new separate `Dockerfile.pydis` that uses the pre-built versions from [`python-discord/python-builds`](https://github.com/python-discord/python-builds).

An integration specific image using this Dockerfile is built for integration tests to mirror the current behaviour.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,24 @@ Name | Description
 `SNEKBOX_DEBUG` | Enable debug logging if set to a non-empty value.
 `SNEKBOX_SENTRY_DSN` | [Data Source Name] for Sentry. Sentry is disabled if left unset.
 
+### Additional Interpreters
+
+By default, snekbox will use and make available the system Python (the version used to run snekbox itself). Additional interpreters or binaries should be placed in `/snekbin/` which is mounted to the nsjail container.
+
+You can use the `executable_path` parameter in `POST` requests to `/eval` to select an interpreter, e.g.
+
+```js
+{
+  "executable_path": "/snekbin/python/3.14/bin/python",
+  // Rest of /eval body
+  ...
+}
+```
+
+The default interpreter is at `/snekbin/python/default/bin/python`, you can symlink `/snekbin/python/default` to another interpreter such as `/snekbin/python/3.14` to change this default.
+
+See [`Dockerfile.pydis`](Dockerfile.pydis) for an example using additional prebuilt interpreters and how to change the defaults.. This uses images built from [`python-discord/python-builds`](https://github.com/python-discord/python-builds).
+
 ## Third-party Packages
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -2,21 +2,26 @@ on:
...
 jobs:
   build:
-    name: Build snekbox-venv image
+    name: Build snekbox-integration image
     runs-on: ubuntu-latest
+    services:
+      registry:
+        image: registry:2
+        ports:
+          - 5000:5000
```
MarkKoz: What's the purpose of hosting a registry in our job?

snekbox/__main__.py
```diff
@@ -13,18 +13,19 @@ def parse_args() -> argparse.Namespace:
     )
     parser.add_argument("code", help="the Python code to evaluate")
     parser.add_argument(
-        "nsjail_args",
-        action="store_const",
-        const=[],
+        "--nsjail-args",
+        nargs="*",
+        default=[],
+        dest="nsjail_args",
```
MarkKoz: What does the help output look like now?

MarkKoz: Why is the default shown twice now?

MarkKoz: They're now options rather than positional arguments. That's confusing.

jb3: > Why is the default shown twice now?

We hard-coded it previously but now switching it from store_const appears to display it again, we can remove our hardcoding.

> They're now options rather than positional arguments. That's confusing.

Yeah, I'll be honest I'd just updated it to maintain functionality because argparse in 3.14 broke it, hadn't looked at how it changed the help command.

I'll have a play around to see if I can maintain behaviour on 3.14 and keep the help command useful.

jb3: Fixed this now:
```
root@038be88be536:/snekbox# snekbox 'some code' --some-nsjail-arg 1234 --- -c abcdef --some-python-arg
Namespace(code='some code', nsjail_args=['--some-nsjail-arg', '1234'], py_args=['-c', 'abcdef', '--some-python-arg'])
```

Usage is now:
```
usage: snekbox [-h] code [nsjail_args ...] [--- py_args ...]

positional arguments:
  code                  the Python code to evaluate
  nsjail_args           override configured NsJail options (default: [])
  py_args               arguments to pass to the Python process (default: ['-c'])

options:
  -h, --help            show this help message and exit
```

Duplicated defaults are also removed. It's a bit cursed but it does work.

